PR #33, opened 2020-12-25
Add ContextualCommandBar component

Yat3s:
### Platforms Impacted
- [x] Android

### Description of changes
Add ContextualCommandBar component in Fluentui-android, it supports
- Config command item and group dynamically
- Customize command item appearance

### Verification
Design spec: https://www.figma.com/file/77nBSZEV9D2lONQwLPjhbK/CCB?node-id=40%3A0
| Light                                       | Dark                                      |
| screenshot-1611906163377| screenshot-1611906135409|

This PR has considered:
- [x] VoiceOver and Keyboard Accessibility
- [x] Internationalization and Right to Left layouts
- [x] Size classes and window sizes (notched devices, multitasking, different window sizes, etc)


Yat3s:
Looks like there is a check failure since I unable to access some resources? 
image

aishwarya56:
@Yat3s Currently build pipeline is in maintenance mode and it is not a mandatory check. You can ignore the failing pipeline

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/DefaultCommandItem.kt
```diff
@@ -0,0 +1,25 @@
+package com.microsoft.fluentui.contextualcommandbar
+
+open class DefaultCommandItem(
+        private val icon: Int,
```
mnkshukl, line 4: @DrawableRes Annotate

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,262 @@
...
+    private var commandContainer = LinearLayout(context)
+
+    // Layout configuration
+    private var groupSpace = 0
+    private var itemSpace = 0
+    private var itemPaddingVertical = 0
+    private var itemPaddingHorizontal = 0
+
+    // Dismiss button configuration
+    var showDismiss = false
```
mnkshukl, line 37: Question:
can we rely upon one ? because I don't see a default dismiss icon.
So  if showDismiss  = true and no Dismiss Icon set - t won't be shown.
So can we make showDismiss based on dismissIcon is set ?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,262 @@
...
+        set(value) {
+            if (value == 0) {
+                return
+            }
+
+            field = value
+            dismissButton?.setImageResource(dismissIcon)
+        }
+    private var dismissButton: ImageView? = null
+    private var dismissButtonContainer: LinearLayout? = null
```
mnkshukl, line 54: nit : private vars on Top

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,262 @@
...
+
+    var dismissListener: (() -> Unit)? = null
+    var itemClickListener: OnItemClickListener? = null
+
+    init {
+        removeAllViews()
+        horizontalScrollView.removeAllViews()
```
mnkshukl, line 61: When is this required ? I See a new instance of horizontalScrollView here and no child views attached

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,262 @@
...
+
+        with(horizontalScrollView) {
+            isVerticalScrollBarEnabled = false
+            isHorizontalScrollBarEnabled = false
+        }
+
+        attrs?.let {
+            val styledAttributes = context.theme.obtainStyledAttributes(
+                    it,
+                    R.styleable.ContextualCommandBar2,
```
mnkshukl, line 78: should be named better, why 2 ?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,262 @@
...
+
+            val groupContainer = LinearLayout(context)
+            commandContainer.addView(groupContainer)
+            if (groupIdx != itemGroups.size - 1) {
+                (groupContainer.layoutParams as LinearLayout.LayoutParams).apply {
+                    marginEnd = groupSpace
+                }
+            }
+
+            for ((idx, item) in items.withIndex()) {
```
mnkshukl, line 199: can we break it to another function  for more readability 
like addItemViewToGroup or similar

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,262 @@
...
+            itemPaddingVertical = resources.getDimensionPixelSize(
+                    R.dimen.fluentui_contextual_command_bar_default_item_padding_vertical
+            )
+            itemPaddingHorizontal = resources.getDimensionPixelSize(
+                    R.dimen.fluentui_contextual_command_bar_default_item_padding_horizontal
+            )
+        }
+
+        // Initialize dismiss button
+        addDismissButton()
```
mnkshukl, line 109: is this needed to be added if visibility is off. Should we not add it only if showVisible is true ?

mnkshukl (commented):
is there a Duo support that needs to be added as well ?
Also about the accessibility

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
Yat3s: Good catch!

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
Yat3s: Yes, make sense!

Yat3s:
@mnkshukl Thanks, added Duo screenshot

For accessibility, I just add `getContentDescription()` for setting `contentDescription` for each CommandItem, does it enough?

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ContextualCommandBarActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+        )
+
+        itemGroups.add(CommandItemGroup()
+                .addItem(DefaultCommandItem(R.drawable.ic_fluent_link_24_regular, "Link", false))
+        )
+
+        with(contextual_command_bar_default) {
+            setItemGroups(itemGroups)
+            itemClickListener = object : ContextualCommandBar.OnItemClickListener {
+                override fun onItemClick(item: CommandItem, view: View) {
```
owenconnolly, line 53: CommandItem has no property for an identifier, so the expectation is that listeners will differentiate by description? If two groups had CommandItems with the same description but for some reason the context should result in a different event, it may be problematic. Is there any reason we should not add an optional string identifier?

FluentUI.Demo/src/main/res/drawable/ic_fluent_add_24_regular.xml
```diff
@@ -0,0 +1,3 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android" android:width="24dp" android:height="24dp" android:viewportWidth="24" android:viewportHeight="24">
```
owenconnolly, line 1: To be clear about the design here, we will expect the consumer to provide the icons, so in order to use the Fluent icons the consumer should provide those independently?

owenconnolly: (and I presume not using the Fluent-System-Icons library instead of adding these resources is because it would add too much to the package size? it looks like these icons are all taken from that library, so any updates to the icons at source will not be reflected without manual updates)

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
...
+import android.view.ViewGroup
+import android.widget.FrameLayout
+import android.widget.HorizontalScrollView
+import android.widget.ImageView
+import android.widget.LinearLayout
+import com.microsoft.fluentui.R
+import com.microsoft.fluentui.util.ThemeUtil
+import com.microsoft.fluentui.util.isVisible
+
+class ContextualCommandBar @JvmOverloads constructor(
```
owenconnolly, line 20: Just starting to look through the PR, and it is not clear to me how icon state is to be updated. Updating the CommandItem selected/enabled state will propagate to the UI representation through the theme bindings?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
...
+                    R.dimen.fluentui_contextual_command_bar_dismiss_button_width
+            ) + resources.getDimensionPixelSize(
+                    R.dimen.fluentui_contextual_command_bar_dismiss_gap_width
+            )
+        } else 0
+        commandContainer.setPaddingRelative(0, 0, paddingEnd, 0)
+        dismissButtonContainer?.isVisible = visible
+    }
+
+    fun setItemGroups(itemGroups: List<CommandItemGroup>) {
```
owenconnolly, line 192: Updating the commands has to be done in a batch operation? We don't expect more granular add/remove icons? I understand the logic to do so would be complex, but wanted to know if it was considered in the design (I think the current design assumes a relatively static set of commands)

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.fluentui.contextualcommandbar
+
+import android.support.annotation.DrawableRes
+
+interface CommandItem {
```
owenconnolly, line 5: should we have doc comments for the classes/methods? it looks like that is the norm for the library code...

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
```diff
@@ -0,0 +1,15 @@
...
+
+    @DrawableRes
+    fun getIcon(): Int
+
+    fun getContentDescription(): String? = null
```
owenconnolly, line 10: is this intended to be a used for accessibility?

owenconnolly: a related question, items will only be rendered as icons, so if there is no icon available the command button can not render (e.g. with a label instead)?
I ask as I have a scenario where commands can be added on the fly, so we may not have an icon to map the command to...

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
owenconnolly, line 5: (also, I added a comment above but will add as it is most relevant here - should we have a property for a command string identifier?)

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItemGroup.kt
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.fluentui.contextualcommandbar
+
+class CommandItemGroup {
```
owenconnolly, line 3: (may not be relevant if we go with a more static CCB design, but should groups also have an identifier property?)

FluentUI/src/main/res/values/themes.xml
```diff
@@ -106,6 +106,17 @@
         <item name="fluentuiCalendarOtherMonthBackgroundColor">@color/fluentui_gray_25</item>
         <item name="fluentuiCalendarSelectedColor">?attr/fluentuiForegroundSelectedColor</item>
         <item name="fluentuiCalendarDayTodayBackgroundColor">?attr/fluentuiColorPrimaryLighter</item>
+
+        <!--ContextualCommandBar-->
+        <item name="fluentuiContextualCommandBarBackgroundColor">@color/fluentui_gray_50</item>
+        <item name="fluentuiContextualCommandBarBackgroundColorPressed">@color/fluentui_gray_100</item>
+        <item name="fluentuiContextualCommandBarBackgroundColorSelected">#eff6fc</item>
```
owenconnolly, line 113: should the custom colors here all be defined Fluent colors? (e.g. #eff6fc etc)

FluentUI/src/main/res/values-night/themes.xml
```diff
@@ -60,6 +60,16 @@
         <item name="fluentuiCalendarDayTextActiveColor">?attr/fluentuiForegroundColor</item>
         <item name="fluentuiCalendarWeekHeadingWeekendTextColor">@color/fluentui_gray_600</item>
 
+        <!--ContextualCommandBar-->
+        <item name="fluentuiContextualCommandBarBackgroundColor">@color/fluentui_gray_600</item>
+        <item name="fluentuiContextualCommandBarBackgroundColorPressed">@color/fluentui_gray_900</item>
+        <item name="fluentuiContextualCommandBarBackgroundColorSelected">#0086F0</item>
```
owenconnolly, line 66: (same for the custom defined colors here, should they be Fluent colors?)

FluentUI.Demo/src/main/res/color/fluent_default_icon_tint.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ Copyright (c) 2019.
```
owenconnolly, line 3: Looks like we should update all of the copyright format in the files being added to match the existing format? (which includes the MIT license line)

owenconnolly: (some files have no copyright header)

FluentUI/src/main/res/drawable/contextual_command_bar_start_item_background.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?><!--
```
owenconnolly, line 1: (nit: newline for copyright block)

FluentUI/src/main/res/drawable/contextual_command_bar_single_item_background.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?><!--
```
owenconnolly, line 1: (also newline here)

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ContextualCommandBarActivity.kt
```diff
@@ -0,0 +1,73 @@
+package com.microsoft.fluentuidemo.demos
```
owenconnolly, line 1: nit: add copyright statement

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ContextualCommandBarActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+    override val contentLayoutId: Int
+        get() = R.layout.activity_contextual_command_bar
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        val itemGroups = arrayListOf<CommandItemGroup>()
+
+        itemGroups.add(CommandItemGroup()
+                .addItem(DefaultCommandItem(R.drawable.ic_fluent_add_24_regular, "Add"))
```
owenconnolly, line 25: should the description values also be localized (if we are localizing 'Click' for the toast below)?

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ContextualCommandBarActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+                }
+            }
+        }
+
+        with(contextual_command_bar_with_dismiss) {
+            setItemGroups(itemGroups)
+            showDismiss = true
+            dismissIcon = R.drawable.ic_fluent_keyboard_dock_24_regular
+            dismissListener = {
+                Toast.makeText(this@ContextualCommandBarActivity, "Dismiss", Toast.LENGTH_SHORT).show()
```
owenconnolly, line 64: (also localize 'Dismiss'?)

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
+package com.microsoft.fluentui.contextualcommandbar
```
owenconnolly, line 1: nit: missing copyright statements

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
owenconnolly: Yep, that would make sense

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
...
+    var itemClickListener: OnItemClickListener? = null
+
+    init {
+        removeAllViews()
+        addView(horizontalScrollView)
+        horizontalScrollView.addView(commandContainer)
+
+        with(commandContainer) {
+            orientation = LinearLayout.HORIZONTAL
+            gravity = Gravity.CENTER
```
owenconnolly, line 65: likely generally the desired layout, but should this be configurable?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
...
+    private fun addDismissButton() {
+        // Container
+        dismissButtonContainer = LinearLayout(context).apply {
+            orientation = LinearLayout.HORIZONTAL
+        }
+        addView(dismissButtonContainer)
+        (dismissButtonContainer!!.layoutParams as LayoutParams).apply {
+            height = ViewGroup.LayoutParams.MATCH_PARENT
+            width = ViewGroup.LayoutParams.WRAP_CONTENT
+            gravity = Gravity.END
```
owenconnolly, line 118: likely no, but do we every envisage a scenario where a consumer might want the dismiss button on the leading side of the command groups (i.e. the left in LTR)?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
...
+            width = resources.getDimensionPixelSize(
+                    R.dimen.fluentui_contextual_command_bar_dismiss_button_width
+            )
+        }
+
+        bringChildToFront(dismissButtonContainer)
+    }
+
+    private fun setDismissButtonVisible(visible: Boolean) {
+        if (dismissIcon == 0) {
```
owenconnolly, line 173: would it make sense to use the setting of the dismiss icon or not to determine whether or not we show dismiss? if set we show it, if not we don't? then we don't need to setShowDismiss and setDismissIcon

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
...
+    private fun addItemViewsToGroup(items: List<CommandItem>, groupContainer: ViewGroup) {
+        for ((idx, item) in items.withIndex()) {
+            val itemView = ImageView(context).apply {
+                setPaddingRelative(
+                        itemPaddingHorizontal,
+                        itemPaddingVertical,
+                        itemPaddingHorizontal,
+                        itemPaddingVertical
+                )
+                setImageResource(item.getIcon())
```
owenconnolly, line 222: if icon is null what do we display?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
...
+                isSelected = item.isSelected() && item.isEnabled()
```
owenconnolly, line 223: is this assumption correct? for example if you have selected text which is all Bold, the Bold command should be selected, but if the content is read only it may also be disabled? read-only content should generally not disable selection, but I am not sure if the design of the CCB UX is that it should reflect context changes in this case...

owenconnolly (changes requested):
I had a few minor comments, and some other questions - some just clarifications, others may require a change (or not), thanks!

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,276 @@
...
+        }
+        dismissButtonContainer!!.addView(dismissButton)
+        dismissButton!!.layoutParams.apply {
+            height = ViewGroup.LayoutParams.MATCH_PARENT
+            width = resources.getDimensionPixelSize(
+                    R.dimen.fluentui_contextual_command_bar_dismiss_button_width
+            )
+        }
+
+        bringChildToFront(dismissButtonContainer)
```
aishwarya56, line 169: Why do we need this?

aishwarya56, line 169: Do we actually need this?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.fluentui.contextualcommandbar
+
+import android.support.annotation.DrawableRes
+
+interface CommandItem {
+
+    @DrawableRes
+    fun getIcon(): Int
```
owenconnolly, line 8: A more general question - I have asked the same on the iOS/Swift implementation PR - it seems there may be scenarios where there is a need to display a command button with a localized string/label rather than an icon. Can we add that functionality here?
I am given to understand that app which hosted the original code this PR is based on does also have that functionality (though I have not checked so am not sure how accurate that is)

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/DefaultCommandItem.kt
```diff
@@ -0,0 +1,27 @@
+package com.microsoft.fluentui.contextualcommandbar
+
+import android.support.annotation.DrawableRes
+
+open class DefaultCommandItem(
```
owenconnolly, line 5: A question on the expectation for updating comment state after some command button is clicked - the default behaviour does not appear to allow for updating state, so I believe the expectation is that consumers will have a custom implementation of CommandItem which will update the internal state as needed, and that should be reflected in the UI after any event?

owenconnolly: On looking at this further, I do not see where the binding to update the UI state is, is there a missing call to update the CommandItem state whenever the data layer changes?

FluentUI.Demo/src/main/res/drawable/ic_fluent_add_24_regular.xml
Yat3s: > To be clear about the design here, we will expect the consumer to provide the icons, so in order to use the Fluent icons the consumer should provide those independently?

We just add these icon resources for demonstration, I think we can delete these and using other icons instead

>(and I presume not using the Fluent-System-Icons library instead of adding these resources is because it would add too much to the package size? it looks like these icons are all taken from that library, so any updates to the icons at source will not be reflected without manual updates)

I think it doesn't increase any package size since these icons are only added in the Demo module instead of the Library module.

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
Yat3s: > is this intended to be a used for accessibility?  

Yes
> a related question, items will only be rendered as icons, so if there is no icon available the command button can not render (e.g. with a label instead)?
I ask as I have a scenario where commands can be added on the fly, so we may not have an icon to map the command to...

Do you mean we need to add an option for display text instead of icon?

owenconnolly: Yes, we have cases where some buttons will not have an icon and we will want them showing just a localized string. This isn't currently possible, could it be added (priority is to show icon, if no icon show display text, if neither icon or display text do not show the button).
Currently if an icon is not provided an empty button is loaded, but probably it should not be?

FluentUI.Demo/src/main/res/drawable/ic_fluent_add_24_regular.xml
owenconnolly: Yes, that is correct. For our case we should include the icon resources with our UI library, rather than have the host do this. We would intend to use icons from the Fluent-System-Icons library also for this case. Does this approach seem correct to you?

FluentUI/src/main/res/values-night/themes.xml
```diff
@@ -63,11 +63,11 @@
         <!--ContextualCommandBar-->
         <item name="fluentuiContextualCommandBarBackgroundColor">@color/fluentui_gray_600</item>
         <item name="fluentuiContextualCommandBarBackgroundColorPressed">@color/fluentui_gray_900</item>
-        <item name="fluentuiContextualCommandBarBackgroundColorSelected">#0086F0</item>
+        <item name="fluentuiContextualCommandBarBackgroundColorSelected">@color/fluentui_communication_blue</item>
         <item name="fluentuiContextualCommandBarIconTint">@color/fluentui_gray_100</item>
         <item name="fluentuiContextualCommandBarIconTintDisabled">@color/fluentui_gray_400</item>
         <item name="fluentuiContextualCommandBarIconTintSelected">@color/fluentui_black</item>
-        <item name="fluentuiContextualCommandBarDismissBackgroundColor">#000000</item>
+        <item name="fluentuiContextualCommandBarDismissBackgroundColor">@color/black</item>
```
owenconnolly, line 70: This causes a compilation/link error for me when pulling in the library, it should be @color/fluentui_black I think?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItemAdapter.kt
```diff
@@ -0,0 +1,170 @@
...
+import android.support.v4.content.ContextCompat
+import android.support.v7.content.res.AppCompatResources
+import android.support.v7.widget.RecyclerView
+import android.view.View
+import android.view.ViewGroup
+import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
+import android.widget.ImageView
+import com.microsoft.fluentui.R
+
+class CommandItemAdapter(val context: Context) : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
```
owenconnolly, line 19: Interestingly because the Adapter is needed in order to set the click event, this requires the consumer to add a dependency on RecyclerView, which may or may not be desired. I am not sure if we want to keep that internal and instead allow the consumer to set some more generic event?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
```diff
@@ -0,0 +1,22 @@
...
+import android.support.annotation.DrawableRes
+
+interface CommandItem {
+
+    fun getId(): String? = null
+
+    @DrawableRes
+    fun getIcon(): Int = 0
+
+    fun getLabel(): String? = null
```
owenconnolly, line 17: We don't still need the original Description string for accessibility context?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
```diff
@@ -0,0 +1,164 @@
...
+        }
+        if (dismissItemGravity == DismissItemGravity.START) {
+            dismissButtonDivider!!.setBackgroundResource(
+                    R.drawable.contextual_command_bar_dismiss_button_divider_start_background
+            )
+            dismissButtonContainer!!.removeAllViews()
+            dismissButtonContainer!!.addView(dismissButton)
+            dismissButtonContainer!!.addView(dismissButtonDivider)
+        } else if (dismissItemGravity == DismissItemGravity.END) {
+            dismissButtonDivider!!.setBackgroundResource(
```
owenconnolly, line 81: JFYI, I am seeing an exception thrown here when I have the Dismiss button displayed at the end and the buttons state updates, kotlin.KotlinNullPointerException, you may be able to repro in the Demo app

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/ContextualCommandBar.kt
Yat3s: image
From the [design spec](https://www.figma.com/file/77nBSZEV9D2lONQwLPjhbK/%5BMobile%5D-CCB?node-id=40%3A0), we should add a small shadow and bring it to the front of the command list

Yat3s: For the above comment, I think this code ensures the dismiss button always on the front of the command list

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
Yat3s: Supported text display

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItemAdapter.kt
Yat3s: Removed the click listener to CommandItem

FluentUI.Demo/src/main/res/drawable/ic_fluent_add_24_regular.xml
Yat3s: Do you mean we should add `com.microsoft.design:fluent-system-icons:version` into our `FluentUI.Demo` instead of importing icon resource?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
```diff
@@ -0,0 +1,51 @@
...
+interface CommandItem {
+
+    interface OnItemClickListener {
+        fun onItemClick(item: CommandItem, view: View)
+    }
+
+    /**
+     * The identifier of this item.
+     */
+    fun getId(): String? = null
```
mnkshukl, line 20: in general context : id is an integer as consumer define it in 
ids.xml file 
can you create 2 variants - one getId - > Int?  and another  getTag ->String ?

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/DefaultCommandItem.kt
owenconnolly: Thanks, I confirmed that state now updates as you would expect

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
owenconnolly: Thanks, that would make sense and should work for all purposes

FluentUI/src/main/java/com/microsoft/fluentui/contextualcommandbar/CommandItem.kt
owenconnolly: Thank you, that works nicely for items with labels now

owenconnolly (approved):

aishwarya56 (approved):

PR #46, opened 2021-02-16
Add secondary custom subtitle view

SurbhiVijayvargeeya:
### Platforms Impacted
- Android (All) - provides extra functionality to add secondary subtitle

### Description of changes

Added an extra custom view in ListItemsView's subtitle

### Verification

Manually tested by running on simulator. Scenarios tested:
1. Portrait mode
2. Landscape mode
3. No changes in previous usage

| 
image
 | 
image
|

### Pull request checklist

This PR has considered:


FluentUI/src/main/res/layout/view_list_item.xml
```diff
@@ -33,14 +34,37 @@
...
-            tools:text="Subtitle"
-            tools:visibility="visible" />
+            android:layout_marginTop="@dimen/fluentui_list_item_text_vertical_spacing">
+
+            <TextView
+                android:id="@+id/list_item_subtitle"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:textAppearance="@style/TextAppearance.FluentUI.ListItemSubtitle"
+                app:layout_constraintTop_toTopOf="@+id/list_item_subtitle_view_container"
```
aishwarya56, line 48: Should we not use, maxLines and ellipsize somewhere? 
Have you tested it for a very long text?

aishwarya56 (commented):
Also, in the screenshot. We see that ellipsize is happening in the start, is it needed?
We should provide an option to customise it if that's the case. Generally it should be in the end

FluentUI/src/main/res/layout/view_list_item.xml
SurbhiVijayvargeeya: This is using previous subtitle only which provides configurable maxLines and ellipsize properties. I have just added one more custom view beside the subtitle.

SurbhiVijayvargeeya: "_ellipsize is happening in the start_" This is just an example. Since it is the same subtitle which was present earlier as well...ellipsize and maxLines are fully configurable.

aishwarya56 (approved):

PR #55, opened 2021-04-01
Androidx movement

aishwarya56:
Shift to Android X
Change target version to 30
Update to latest Kotlin and gradle version
Remove the use of kotlinx.synthetic as it is deprecated now. using viewbinding wherever possible
Here is the migration guide


FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/AppBarLayoutActivity.kt
```diff
@@ -128,16 +128,16 @@ class AppBarLayoutActivity : DemoActivity(), View.OnClickListener {
...
         super.onSaveInstanceState(outState)
 
         searchbarQuery = searchbar.query.toString()
 
-        outState?.putInt(SCROLL_BEHAVIOR, scrollBehavior.ordinal)
-        outState?.putInt(NAVIGATION_ICON_TYPE, navigationIconType.ordinal)
-        outState?.putBoolean(SEARCHBAR_IS_ACTION_MENU_VIEW, searchbarIsActionMenuView)
-        outState?.putBoolean(SEARCHBAR_HAS_FOCUS, searchbarHasFocus)
-        outState?.putString(SEARCHBAR_QUERY, searchbarQuery)
+        outState.putInt(SCROLL_BEHAVIOR, scrollBehavior.ordinal)
```
msgharpu, line 136: Any specific reason for this change?

fluentui_ccb/build.gradle
```diff
@@ -38,53 +37,54 @@ dependencies {
...
 
-
-publishing {
-    repositories {
-        maven {
-            url 'https://pkgs.dev.azure.com/microsoftdesign/fluentui-native/_packaging/fluentui-android/maven/v1'
-            credentials {
-                username = project.hasProperty("mavenUserName") ? "$mavenUserName" : ""
-                password = project.hasProperty("mavenPassword") ? "$mavenPassword" : ""
+project.afterEvaluate {
```
msgharpu, line 52: Is project.afterEvaluate necessary? Last time we had removed it from modularization PR

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/AppBarLayoutActivity.kt
aishwarya56: Definition of these have changed in androidx , as you can see in the bundle is not nullable now and hence we are overriding correct definition now

fluentui_ccb/build.gradle
aishwarya56: Yes it is necessary. When upgrading from 4.x to 5.x we cannot have publishing directly
here is a gradle doc. Please refer to breaking changes. It is mentioned that we need to move publishing block inside afterEvaluate
https://docs.gradle.org/current/userguide/upgrading_version_4.html

fluentui_drawer/src/main/java/com/microsoft/fluentui/drawer/DrawerDialog.kt
```diff
@@ -59,6 +57,8 @@ open class DrawerDialog @JvmOverloads constructor(context: Context, val behavior
 
     private var isExpanded: Boolean = false
     protected val container: View
+    protected val drawerContent: ViewGroup
+    private val drawer: View
```
msgharpu, line 61: Assuming these changes are unrelated to AndroidX update, right?

fluentui_core/src/main/java/com/microsoft/fluentui/view/BaseDividerItemDecoration.kt
```diff
@@ -8,15 +8,16 @@ package com.microsoft.fluentui.view
...
+import androidx.core.content.ContextCompat
+import androidx.recyclerview.widget.DividerItemDecoration
 import android.view.View
 import com.microsoft.fluentui.core.R
 import com.microsoft.fluentui.theming.FluentUIContextThemeWrapper
 import com.microsoft.fluentui.util.ThemeUtil
 
-open class BaseDividerItemDecoration : DividerItemDecoration {
+open class BaseDividerItemDecoration :
+    androidx.recyclerview.widget.DividerItemDecoration {
```
msgharpu, line 20: nit : Why the DividerItemDecoration has to be fully qualified with package name even if the pkg is imported?

fluentui_drawer/src/main/java/com/microsoft/fluentui/drawer/DrawerDialog.kt
aishwarya56: These are actually, as we cannot reference views directly now and it is deprecated. For reference https://developer.android.com/topic/libraries/view-binding/migration

fluentui_core/src/main/java/com/microsoft/fluentui/view/BaseDividerItemDecoration.kt
aishwarya56: Got missed while using auto-import, got it changed at man other places but missed here. Might be some other places too, will go through it once more

msgharpu (approved):

PR #58, opened 2021-04-12
Use ContextThemeWrapper for others module

msgharpu:
### Platforms Impacted
- [ * ] Android

### Description of changes

This PR adds the context theme wrapper to others module as well. It will avoid the issues for existing users
who are using the complete library when they switch to modular fluent

### Verification

Tested using FluentUI demo app by changing the AppTheme - where the theme doesn't inherit Fluent theme

### Pull request checklist

This PR has considered:
- [* ] Light and Dark appearances
- [ *] VoiceOver and Keyboard Accessibility
- [* ] Internationalization and Right to Left layouts
- [ *] Size classes and window sizes (notched devices, multitasking, different window sizes, etc)

### DIFF
--- a/fluentui_others/src/main/java/com/microsoft/fluentui/actionbar/ActionBarLayout.kt
+++ b/fluentui_others/src/main/java/com/microsoft/fluentui/actionbar/ActionBarLayout.kt
@@ -13,7 +13,7 @@ import kotlinx.android.synthetic.main.view_action_bar.view.*
 import java.lang.Exception
 import java.lang.IllegalStateException
 
-class ActionBarLayout @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) : TemplateView(FluentUIContextThemeWrapper(context), attrs) {
+class ActionBarLayout @JvmOverloads constructor(appContext: Context, attrs: AttributeSet? = null) : TemplateView(FluentUIContextThemeWrapper(appContext,R.style.Theme_FluentUI_Components), attrs) {
     companion object {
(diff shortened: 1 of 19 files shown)


fluentui_others/src/main/java/com/microsoft/fluentui/actionbar/ActionBarLayout.kt
```diff
@@ -13,7 +13,7 @@ import kotlinx.android.synthetic.main.view_action_bar.view.*
 import java.lang.Exception
 import java.lang.IllegalStateException
 
-class ActionBarLayout @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) : TemplateView(FluentUIContextThemeWrapper(context), attrs) {
+class ActionBarLayout @JvmOverloads constructor(appContext: Context, attrs: AttributeSet? = null) : TemplateView(FluentUIContextThemeWrapper(appContext,R.style.Theme_FluentUI_Components), attrs) {
```
aishwarya56, line 16: cant we have default param in FluentUIContextThemeWrapper ? All the constructors seems to pass R.style.Theme_FluentUI_Components

msgharpu: This is only for the others module. As the components are migrated to their own modules, R.style.Theme_FluentUI_Components will start reducing and eventually it won't be there. Hence I think its not required to make it default

aishwarya56 (approved):

PR #71, opened 2021-05-06
Accessibility Bug Fixes related to ActionBarLayout, BottomNavigation

tamasane:
### Platforms Impacted
- [x] Android

### Description of changes

Fixes accessibility bugs in demo app 
- ([AB#10436](https://dev.azure.com/microsoftdesign/e0256516-4c8d-4265-8edc-b3309388a503/_workitems/edit/10436)) ActionBar Demo has radio groups selections for position and type. The label was not announced when navigated using keyboard Tab or swipe gesture. Added contentDescription to RadioGroups. This is now fixed and correctly announces list "radio group" for the first "radio button".
- ([AB#9409](https://dev.azure.com/microsoftdesign/e0256516-4c8d-4265-8edc-b3309388a503/_workitems/edit/9409)) BottomNavigation has Toggle Labels button that toggles the label visibility. The label state was not announced after button is clicked. This is now fixed and Labels are ON/OFF is announced correctly 
- ([AB#9410](https://dev.azure.com/microsoftdesign/e0256516-4c8d-4265-8edc-b3309388a503/_workitems/edit/9410)) BottomNavigation has Buttons to change the number of menu items. The button selection was not announced after button is clicked. This is now fixed and 'Selected' announces correctly 

### Verification

| 10436                                       | 9409                                      | 9410                                      |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|
|  |  |  |

### Pull request checklist

This PR has considered:
- [x] VoiceOver and Keyboard Accessibility

### DIFF
--- a/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ActionBarLayoutActivity.kt
+++ b/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ActionBarLayoutActivity.kt
@@ -2,8 +2,6 @@ package com.microsoft.fluentuidemo.demos
 
 import android.content.Intent
 import android.os.Bundle
-import android.widget.LinearLayout
-import android.widget.TextView
 import com.microsoft.fluentui.actionbar.ActionBarLayout
 import com.microsoft.fluentui.snackbar.Snackbar
 import com.microsoft.fluentui.util.isAccessibilityEnabled
--- a/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ActionBarLayoutActivity.kt
+++ b/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ActionBarLayoutActivity.kt
@@ -2,8 +2,11 @@ package com.microsoft.fluentuidemo.demos
 
 import android.content.Intent
 import android.os.Bundle
+import android.widget.LinearLayout
+import android.widget.TextView
 import com.microsoft.fluentui.actionbar.ActionBarLayout
 import com.microsoft.fluentui.snackbar.Snackbar
+import com.microsoft.fluentui.util.isAccessibilityEnabled
 import com.microsoft.fluentuidemo.Demo
 import com.microsoft.fluentuidemo.DemoActivity
 import com.microsoft.fluentuidemo.R
@@ -17,10 +20,19 @@ class ActionBarLayoutActivity: DemoActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        if(isAccessibilityEnabled){
+            label_actionbar_position.isFocusable = true
+            label_actionbar_type.isFocusable = true
+        }
+        else {
+            label_actionbar_position.isFocusable = false
+            label_actionbar_type.isFocusable = false
+        }
+
         start_demo_btn.setOnClickListener{
             var position = action_bar_position_rgroup.checkedRadioButtonId
             var type = action_bar_type_rgroup.checkedRadioButtonId
-
             if(position == -1 || type == -1){
                 Snackbar.make(root_view,"Please select position and type", Snackbar.LENGTH_SHORT).show()
             }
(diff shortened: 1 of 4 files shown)


FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/BottomNavigationActivity.kt
```diff
@@ -23,17 +23,23 @@ class BottomNavigationActivity : DemoActivity() {
...
-                    LabelVisibilityMode.LABEL_VISIBILITY_LABELED
-                else
-                    LabelVisibilityMode.LABEL_VISIBILITY_UNLABELED
+            if (bottom_navigation.labelVisibilityMode == LabelVisibilityMode.LABEL_VISIBILITY_UNLABELED){
+                bottom_navigation.labelVisibilityMode = LabelVisibilityMode.LABEL_VISIBILITY_LABELED
+                labelsState = "ON"
+            }
+            else {
+                bottom_navigation.labelVisibilityMode = LabelVisibilityMode.LABEL_VISIBILITY_UNLABELED
+                labelsState = "OFF"
```
mnkshukl, line 34: All caps for announcement can pronounce  'O' 'F' 'F' rather than Off , can you verify this in <10 android as well or just prefer word case 'Off' & 'on'

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ActionBarLayoutActivity.kt
```diff
@@ -17,10 +18,19 @@ class ActionBarLayoutActivity: DemoActivity() {
...
+
+        if(isAccessibilityEnabled){
+            label_actionbar_position.isFocusable = true
+            label_actionbar_type.isFocusable = true
+        }
+        else {
+            label_actionbar_position.isFocusable = false
+            label_actionbar_type.isFocusable = false
+        }
+
```
mnkshukl, line 30: @aishwarya56 
I feel mixing of keyboard and accessibility is something across few issues I see.And I don't feel they should be related , Like navigating accessibility control with tab is not a situation , Do you feel these issues should be pushed back ?

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/BottomNavigationActivity.kt
tamasane: It announced correctly on Android 9.0, but changed them to sentence case as mentioned here.

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ActionBarLayoutActivity.kt
aishwarya56: @mnkshukl ,
This check makes sense in some of the cases. Like this AB#9423, here the click on label works only in accessible mode and hence it needs to be announced only when accessibility is enabled and hence the check was required.

But here, the bug is different and it does not require this check.
So, to sum it up, the fix is not correct and it should announce the radio group name along the radio label.

Announcement should be something like,  checked/unchecked "radio label" button in "radio group"  ......
Hence, we do not need this check and the label need not be in focus.

I hope, this makes sense?

tamasane: @aishwarya56 
I thought adding the "radio group" to the announcements of each of the radio button would make the announcement unnecessarily longer, thats why resorted to just making the labels focusable so that they are announced before the radio buttons.

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/ActionBarLayoutActivity.kt
```diff
@@ -4,6 +4,7 @@ import android.content.Intent
 import android.os.Bundle
 import com.microsoft.fluentui.actionbar.ActionBarLayout
 import com.microsoft.fluentui.snackbar.Snackbar
+import com.microsoft.fluentui.util.isAccessibilityEnabled
```
aishwarya56, line 7: We can remove this import

aishwarya56 (approved):

PR #76, opened 2021-05-27
Showcase all scenarios in theming for modularized app

aishwarya56:
The changes are majorly in Fluent Demo app and it showcases how consumer apps can use us without facing difficulties

### DIFF
--- a/FluentUI.Demo/src/main/res/values/styles.xml
+++ b/FluentUI.Demo/src/main/res/values/styles.xml
@@ -6,10 +6,112 @@
 
 <resources xmlns:tools="http://schemas.android.com/tools">
 
-    <style name="Base.AppTheme" parent="Theme.FluentUI">
+    <!-- Example theme with Android parent theme, you will have to override styles and attrs -->
+    <style name="Base.AppTheme" parent="Theme.AppCompat.Light.DarkActionBar">
+        <item name="colorPrimary">@color/fluentui_communication_blue</item>
+        <item name="colorPrimaryDark">@color/fluentui_communication_shade_20</item>
+        <item name="colorAccent">@color/fluentui_communication_blue</item>
+        <item name="android:textColorPrimary">@color/fluentui_gray_900</item>
         <item name="windowActionBar">false</item>
         <item name="windowNoTitle">true</item>
-        <item name="android:windowBackground">?attr/fluentuiBackgroundColor</item>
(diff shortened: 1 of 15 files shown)


msgharpu (approved):

FluentUI.Demo/src/main/res/values/styles.xml
```diff
@@ -6,10 +6,112 @@
...
+                     toolbar title color to red given you have set the style
+             3) Use Style Defined in Library but using your own Widget - Not Recommended. But still
+                    if you decide to do this.
+                    Example - You have set Fluent Library Style
+                    <item name="buttonStyle">@style/Widget.FluentUI.Button</item>
+                    But in app you are not using FluentUI.Widget.Button and directly consuming
+                    Button provided by android. So, you will face inconsistencies in color and
+                    other attributes. To not face inconsistencies you will have to declare the
+                    attrs defined in Library again.
+                    <item name="fluentuiButtonBackgroundDefaultColor">your_color</item>
```
mnkshukl, line 54: Question : What's the way to know all available attribute's name I want to override  in 2nd & 3rd case ?

FluentUI.Demo/src/main/res/values/styles.xml
```diff
@@ -6,10 +6,112 @@
...
+        <!--Launch Screen: Remove the background that was set in the launch screen theme-->
+        <item name="android:background">@null</item>
+        <item name="android:statusBarColor">?android:colorPrimary</item>
+        <item name="android:windowLightStatusBar" tools:targetApi="23">false</item>
+
+
+        <!-- If you decide to use any of the attributes defined in Fluent Library in the demo app ,
+             then those attrs should be defined in your theme or they should be wrapped in
+             FluentUIContextWrapper if using programmatically -->
+        <item name="fluentuiBackgroundColor">@color/fluentui_white</item>
```
mnkshukl, line 28: Should the consumer app override fleuntui.Base.Theme  if it wants to use these attrs in their app, this way it should have basic fluent attrs without having to define them and add the one required ? Otherwise I can't see a case where app would/should do this

aishwarya56: yes the consumer can override fleuntui.Base.Theme this example is shown too in the files and in that case they do not need to declare them if they are using it but only the ones that are present in this theme. After modularization we have multiple themes. So, you need to be sure if that attribute is defined or not

The case we have shown in this example is when you use custom theme and not fluent theme

PR #82, opened 2021-06-18
Accessibility Fixes in persona Module

aishwarya56:
Add contentDescription String to IAvatar interface
This allows consumers to set contentDescription for the views in Persona Module

### DIFF
--- a/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/AvatarViewActivity.kt
+++ b/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/AvatarViewActivity.kt
@@ -58,11 +58,12 @@ class AvatarViewActivity : DemoActivity() {
     }
 
     private fun createNewAvatarFromCode() {
+        val avatarName = getString(R.string.persona_name_mauricio_august)
         val avatarView = AvatarView(this)
         avatarView.avatarSize = AvatarSize.XXLARGE
-        avatarView.name = getString(R.string.persona_name_mauricio_august)
+        avatarView.name = avatarName
         avatarView.email = getString(R.string.persona_email_mauricio_august)
-        avatarView.avatarContentDescriptionLabel  =  getString(R.string.persona_name_mauricio_august)
+        avatarView.avatarContentDescriptionLabel = avatarName
         avatar_circle_example_xxlarge.addView(avatarView)
     }
 }
--- a/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/AvatarViewActivity.kt
+++ b/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/AvatarViewActivity.kt
@@ -62,6 +62,7 @@ class AvatarViewActivity : DemoActivity() {
         avatarView.avatarSize = AvatarSize.XXLARGE
         avatarView.name = getString(R.string.persona_name_mauricio_august)
         avatarView.email = getString(R.string.persona_email_mauricio_august)
+        avatarView.avatarContentDescriptionLabel  =  getString(R.string.persona_name_mauricio_august)
         avatar_circle_example_xxlarge.addView(avatarView)
     }
 }
--- a/fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
+++ b/fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
@@ -15,7 +15,6 @@ import androidx.annotation.ColorInt
 import androidx.core.content.ContextCompat
 import androidx.appcompat.widget.AppCompatImageView
 import android.util.AttributeSet
-import com.microsoft.fluentui.persona.R
 import com.microsoft.fluentui.theming.FluentUIContextThemeWrapper
 
 enum class AvatarStyle {
@@ -146,6 +145,20 @@ open class AvatarView : AppCompatImageView {
             field = value
             initials.setInfo(name, email, avatarBackgroundColor, true)
             invalidate()
+        }
+
+    /**
+     * Defines the [AvatarIsOverFlow] applied to the avatar.
+     */
+    var avatarContentDescriptionLabel: String = ""
+        set(value) {
+            if (field == value)
+                return
+
+            field = value
+            if(avatarContentDescriptionLabel.isNotEmpty()) {
+                contentDescription = avatarContentDescriptionLabel
+            }
         }
 
     private val initials = InitialsDrawable(context)
@@ -308,4 +321,5 @@ fun AvatarView.setAvatar(avatar: IAvatar) {
     avatarImageResourceId = avatar.avatarImageResourceId
     avatarImageUri = avatar.avatarImageUri
     avatarBackgroundColor = avatar.avatarBackgroundColor
+    avatarContentDescriptionLabel  = avatar.avatarContentDescriptionLabel
 }
--- a/fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
+++ b/fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
@@ -156,9 +156,7 @@ open class AvatarView : AppCompatImageView {
                 return
 
             field = value
-            if(avatarContentDescriptionLabel.isNotEmpty()) {
-                contentDescription = avatarContentDescriptionLabel
-            }
+            contentDescription = avatarContentDescriptionLabel
         }
 
     private val initials = InitialsDrawable(context)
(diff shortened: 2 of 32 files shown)


FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/AvatarViewActivity.kt
```diff
@@ -62,6 +62,7 @@ class AvatarViewActivity : DemoActivity() {
         avatarView.avatarSize = AvatarSize.XXLARGE
         avatarView.name = getString(R.string.persona_name_mauricio_august)
```
owenconnolly, line 63: is it any more efficient to store the name in a local var and re-use for both cases? (and in cases below)

aishwarya56: Yea, it will be a bit faster. Incorporating the change

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/IPersona.kt
```diff
@@ -23,4 +23,5 @@ data class Persona(override var name: String = "", override var email: String =
     override var avatarImageResourceId: Int? = null
     override var avatarImageUri: Uri? = null
     override var avatarBackgroundColor: Int? = null
+    override var avatarContentDescriptionLabel: String =""
```
owenconnolly, line 26: (nit: space after =)

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaView.kt
```diff
@@ -151,6 +162,7 @@ class PersonaView : ListItemView {
         avatarView.avatarImageBitmap = avatarImageBitmap
         avatarView.avatarImageUri = avatarImageUri
         avatarView.avatarBackgroundColor = avatarBackgroundColor
+        avatarView.avatarContentDescriptionLabel  = avatarContentDescriptionLabel
```
owenconnolly, line 165: (nit: extra space before =)

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaView.kt
```diff
@@ -113,6 +113,17 @@ class PersonaView : ListItemView {
...
         }
 
+    var avatarContentDescriptionLabel: String = ""
+        set(value) {
+            if (field == value)
+                return
+            field = value
+            avatarView?.apply {
+                avatarContentDescriptionLabel = this@PersonaView.avatarContentDescriptionLabel
+                isFocusable = true
```
owenconnolly, line 123: if we set it to null or empty string should isFocusable = false?

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaChipView.kt
```diff
@@ -80,6 +78,17 @@ class PersonaChipView : TemplateView {
...
         }
+
+    var avatarContentDescriptionLabel: String = ""
+        set(value) {
+            if (field == value)
+                return
+            field = value
+            avatarView?.apply {
+                avatarContentDescriptionLabel = this@PersonaChipView.avatarContentDescriptionLabel
+                isFocusable = true
```
owenconnolly, line 89: (same for this case, should isFocusable = false if the value is null or empty?)

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaChipView.kt
```diff
@@ -214,6 +223,10 @@ class PersonaChipView : TemplateView {
             avatarImageBitmap = this@PersonaChipView.avatarImageBitmap
             avatarImageUri = this@PersonaChipView.avatarImageUri
             avatarBackgroundColor  = this@PersonaChipView.avatarBackgroundColor
+            if(this@PersonaChipView.avatarContentDescriptionLabel.isNotEmpty()) {
```
owenconnolly, line 226: should we be allowed to set it to empty also?

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
```diff
@@ -308,4 +321,5 @@ fun AvatarView.setAvatar(avatar: IAvatar) {
     avatarImageResourceId = avatar.avatarImageResourceId
     avatarImageUri = avatar.avatarImageUri
     avatarBackgroundColor = avatar.avatarBackgroundColor
+    avatarContentDescriptionLabel  = avatar.avatarContentDescriptionLabel
```
owenconnolly, line 324: (nit: extra space before the =)

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
```diff
@@ -148,6 +147,20 @@ open class AvatarView : AppCompatImageView {
             invalidate()
         }
 
+    /**
+     * Defines the [ContentDescription] applied to the avatar.
+     */
+    var avatarContentDescriptionLabel: String = ""
+        set(value) {
+            if (field == value)
```
owenconnolly, line 155: just a question, the repo doesn't apply a style requirement where all if statements require braces {}, even for single line cases?

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
```diff
@@ -148,6 +147,20 @@ open class AvatarView : AppCompatImageView {
...
+                return
+
+            field = value
+            if(avatarContentDescriptionLabel.isNotEmpty()) {
```
owenconnolly, line 159: (same Q here ref should we be able to set it to empty)

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
```diff
@@ -148,6 +147,20 @@ open class AvatarView : AppCompatImageView {
             invalidate()
         }
 
+    /**
+     * Defines the [ContentDescription] applied to the avatar.
+     */
+    var avatarContentDescriptionLabel: String = ""
```
owenconnolly, line 153: will this still read the control type out as button to signify that it is actionable? I presume it will...

owenconnolly (approved):

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
aishwarya56: When empty, it announces as unlabelled something. So, thought not to allow that.

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaView.kt
aishwarya56: Thanks for pointing these outs. My laptop's keyboard is giving some issues while entering spaces and hence these issues.

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
aishwarya56: It is being followed at other places but not in these setvalue overrides across the app. Will have to take it up, in some other refactoring. Right now leaving it as it is

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
aishwarya56: Yea, it will announce it as button. If we have a clicklistener assigned to it. if not, it will just announce the label

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaChipView.kt
aishwarya56: makes sense

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaChipView.kt
aishwarya56: We will. But since it will not be focusable if not empty. it does not actually make sense but we will remove this extra check

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaChipView.kt
owenconnolly: thanks, if it makes sense!

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/PersonaChipView.kt
owenconnolly: thanks, it makes sense that it will not be focusable when empty, that sounds good!

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
owenconnolly: Okay, thanks, I agree with you we should not want it to announce unlabeled when empty, we should definitely avoid that!

fluentui_persona/src/main/java/com/microsoft/fluentui/persona/AvatarView.kt
owenconnolly: great, thanks for confirming!

PR #115, opened 2021-10-20
People picker view fixes

saurabh0802chd:
Fixes remove span announcement
Fixes view rendering when allowCollapse is false
Added APIs to add and remove span
Fixes extra padding if label is blank

### Platforms Impacted
- [X] Android

### Description of changes

PeoplePickerView.kt
- Added APIs to add and remove span
- Hide the labelTextView if label is not provided

PeoplePickerTextView..kt
- Since removeObject post task to handler, added map for remove span notifications
- Added allowCollapse check inside performCollapseAndAdjustLayout method

### Verification

### Pull request checklist

This PR has considered:
- [X] Light and Dark appearances
- [X] VoiceOver and Keyboard Accessibility
- [X] Internationalization and Right to Left layouts
- [X] Size classes and window sizes (notched devices, multitasking, different window sizes, etc)

### DIFF
--- a/fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
+++ b/fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
@@ -170,7 +170,7 @@ internal class PeoplePickerTextView : TokenCompleteTextView<IPersona> {
             else
                 unblockInput()
         }
-    private var shouldAnnouncePersonaAddition: Boolean = false
+    private var shouldAnnouncePersonaAdditionMap = mutableMapOf<IPersona, Boolean>()
     private var shouldAnnouncePersonaRemovalMap = mutableMapOf<IPersona, Boolean>()
     private var searchConstraint: CharSequence = ""
     private var lastSpan: TokenImageSpan? = null
@@ -302,7 +302,9 @@ internal class PeoplePickerTextView : TokenCompleteTextView<IPersona> {
         if (objects.size == personaChipLimit)
             return
 
-        shouldAnnouncePersonaAddition = true
+        lastSpan?.let {
+            shouldAnnouncePersonaAdditionMap[it.token] = true
+        }
         super.replaceText(text)
         context.activity?.let {
             if (DuoSupportUtils.isDualScreenMode(it) && lastSpan != null) {
@@ -427,11 +429,17 @@ internal class PeoplePickerTextView : TokenCompleteTextView<IPersona> {
         }
     }
 
+    /**
+     * Add a picked persona
+     */
     fun addPickedPersona(persona: IPersona) {
-        shouldAnnouncePersonaAddition = true
+        shouldAnnouncePersonaAdditionMap[persona] = true
         super.addObject(persona)
     }
 
+    /**
+     * Removes a persona from picked items
+     */
     fun removePickedPersona(persona: IPersona) {
         shouldAnnouncePersonaRemovalMap[persona] = true
         super.removeObject(persona)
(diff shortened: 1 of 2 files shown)


fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
```diff
@@ -612,18 +632,20 @@ internal class PeoplePickerTextView : TokenCompleteTextView<IPersona> {
 
     private class TokenListener(val view: PeoplePickerTextView) : TokenCompleteTextView.TokenListener<IPersona> {
         override fun onTokenAdded(token: IPersona) {
-            if (view.shouldAnnouncePersonaAddition)
+            if (view.shouldAnnouncePersonaAdditionMap[token] != false)
```
aishwarya56, line 635: Where are we setting this to true?

fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
```diff
@@ -297,7 +302,6 @@ internal class PeoplePickerTextView : TokenCompleteTextView<IPersona> {
         if (objects.size == personaChipLimit)
             return
 
-        shouldAnnouncePersonaAddition = true
```
aishwarya56, line 300: Do we need this?

fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
```diff
@@ -308,13 +312,14 @@ internal class PeoplePickerTextView : TokenCompleteTextView<IPersona> {
 
     override fun canDeleteSelection(beforeLength: Int): Boolean {
         // This method is called from keyboard events so any token removed would be coming from the user.
-        shouldAnnouncePersonaRemoval = true
```
aishwarya56, line 311: Same here?

fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
```diff
@@ -612,18 +632,20 @@ internal class PeoplePickerTextView : TokenCompleteTextView<IPersona> {
...
                 view.tokenListener?.onTokenAdded(token)
             if (view.isFocused)
                 view.announcePersonaAdded(token)
             view.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUS_CLEARED)
+            view.shouldAnnouncePersonaAdditionMap.remove(token)
         }
 
         override fun onTokenRemoved(token: IPersona) {
-            if (view.shouldAnnouncePersonaRemoval)
+            if (view.shouldAnnouncePersonaRemovalMap[token] != false)
```
aishwarya56, line 644: cant we directly do view.shouldAnnouncePersonaRemovalMap[token] == true
otherwise it will work for null also

saurabh0802chd: By default we will be announcing the persona removal and addition. If in any case we don't need the notification we have to set it to false in the corresponding map.

fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
saurabh0802chd: Since announcement of span removal and addition is on by default, so setting it to true is redundant.

fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
saurabh0802chd: Since announcement of span removal and addition is on by default, so setting it to true is redundant.

fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/peoplepicker/PeoplePickerTextView.kt
saurabh0802chd: We don't need to, as announcements are ON by default.

aishwarya56 (approved):

PR #116, opened 2021-10-27
Tooltip Orientation and Second Screen Fixes

amitavsonims:
### Platforms Impacted
- [X] Android

### Description of changes

1. Changed anchor.getLocationOnScreen(screenPos) to anchor.getLocationInWindow(screenPos). Both of these return the same values in most conditions. This changes helps in solving shifts(due to status bar) when device is rotated 90 ACW. 
2. Removed hasSoftNavigationBar check on device 270 ACW rotation. We get a 0 navigationBar height if it is not present. Removed the comment that this check does not work on emulator.
3. Assuming views going off-screen as a second screen scenario is incorrect as views may be defined that cross device boundaries. Added an explicit check for second screen in these scenarios.
4. OnDismissListener was not being called on Dismiss by touching outside popup window, corrected.

### Verification

Changes verified manually on Pixel4 API 30 in 2 configurations: Navigation Bar On, Navigation Bar Off(Gesture Navigation mode)

### Pull request checklist

This PR has considered:
- [X] Light and Dark appearances
- [X] VoiceOver and Keyboard Accessibility
- [X] Internationalization and Right to Left layouts
- [X] Size classes and window sizes (notched devices, multitasking, different window sizes, etc)

### DIFF
--- a/fluentui_core/src/main/java/com/microsoft/fluentui/util/DisplayUtils.kt
+++ b/fluentui_core/src/main/java/com/microsoft/fluentui/util/DisplayUtils.kt
@@ -41,7 +41,7 @@ val Context.navigationBarHeight: Int
 val Context.softNavBarOffsetX: Int
     get() {
         val rotation = (getSystemService(Context.WINDOW_SERVICE) as WindowManager).defaultDisplay.rotation
-        return if (rotation == Surface.ROTATION_270 && hasSoftNavigationBar && !isTablet)
+        return if (rotation == Surface.ROTATION_270 && !isTablet)
             navigationBarHeight
         else
             0
@@ -66,13 +66,6 @@ val Context.desiredDialogSize: IntArray
         dialogSize[1] = WindowManager.LayoutParams.WRAP_CONTENT
 
         return dialogSize
-    }
-
-// This check returns false in emulators
-val Context.hasSoftNavigationBar: Boolean
-    get() {
-        val id = resources.getIdentifier(CONFIG_SHOW_NAVIGATION_BAR, BOOL_STRING, ANDROID_STRING)
-        return id > 0 && resources.getBoolean(id)
     }
 
 val Context.displaySize: Point
--- a/fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
+++ b/fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
@@ -147,15 +147,13 @@ class Tooltip {
         if (requireReinit ) initTooltipArrow(anchorRect, anchor.layoutIsRtl, config.offsetX)
         if (requireReadjustment) readjustTooltip(anchorRect, anchor.layoutIsRtl, config)
 
-        if (config.touchDismissLocation == TouchDismissLocation.INSIDE) {
-            tooltipView.x = if (anchor.layoutIsRtl) resetPositionXForRtl() else positionX.toFloat()
-            tooltipView.y = positionY.toFloat()
-            anchor.post { popupWindow.showAtLocation(anchor, Gravity.NO_GRAVITY, 0, 0) }
-        } else {
-            popupWindow.width = contentWidth
-            popupWindow.height = contentHeight
-            anchor.post { popupWindow.showAtLocation(anchor, Gravity.NO_GRAVITY, positionX, positionY)}
-        }
+        popupWindow.width = contentWidth
+        popupWindow.height = contentHeight
+        anchor.post { popupWindow.showAtLocation(anchor, Gravity.NO_GRAVITY, positionX, positionY)}
+
+        if (config.touchDismissLocation == TouchDismissLocation.INSIDE)
+            popupWindow.setOutsideTouchable(false);
+
         // popupWindow may get dismissed by outside touch for TouchDismissLocation.ANYWHERE
         popupWindow.setOnDismissListener {
             dismissSideEffects()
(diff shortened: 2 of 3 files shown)


fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
```diff
@@ -132,7 +132,7 @@ class Tooltip {
 
         // Get location of anchor view on screen
         val screenPos = IntArray(2)
-        anchor.getLocationOnScreen(screenPos)
+        anchor.getLocationInWindow(screenPos)
```
aishwarya56, line 135: Why do we need this? All calculations are done as per screen right?
Also, how will showAtlocation behave if we are getting location with window reference?

fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
```diff
@@ -173,12 +177,15 @@ class Tooltip {
...
 
-    fun dismiss() {
+    private fun dismissSideEffects() {
         tooltipView.announceForAccessibility(context.getString(R.string.tooltip_accessibility_dismiss_announcement))
-        popupWindow.dismiss()
         onDismissListener?.onDismiss()
     }
 
+    fun dismiss() {
+        popupWindow.dismiss()
```
aishwarya56, line 186: we will have to call listener from here too

fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
```diff
@@ -295,12 +302,12 @@ class Tooltip {
         val startPosition = positionX + layoutParams.marginStart
         val topBarHeight = context.statusBarHeight + ( context.activity?.supportActionBar?.height ?: 0 )
         val doesNotFitAboveOrBelow = (positionY < topBarHeight) || (positionY + contentHeight > displayHeight)
-        val rightSpace = displayWidth - anchorRect.right - context.softNavBarOffsetX
-        val rightEdge = ( startPosition + upArrowWidth + cornerRadius + margin > displayWidth ) || (doesNotFitAboveOrBelow && anchorRect.left > rightSpace)
+        val rightSpace = displayWidth - anchorRect.right + context.softNavBarOffsetX
```
aishwarya56, line 305: Why are we changing this?
If we are calculating it from right and the navbar is shown in right then it should be subtracted?

fluentui_core/src/main/java/com/microsoft/fluentui/util/DisplayUtils.kt
```diff
@@ -41,7 +41,7 @@ val Context.navigationBarHeight: Int
 val Context.softNavBarOffsetX: Int
     get() {
         val rotation = (getSystemService(Context.WINDOW_SERVICE) as WindowManager).defaultDisplay.rotation
-        return if (rotation == Surface.ROTATION_270 && hasSoftNavigationBar && !isTablet)
```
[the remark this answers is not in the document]
amitavsonims: As the comment there states, this check is faulty on emulators. In my case, I found the check faulty on Pixel 5 emulator as well as physical device. 
Yes we get 0 when navbar is disabled.

fluentui_core/src/main/java/com/microsoft/fluentui/util/DisplayUtils.kt
```diff
@@ -68,13 +68,6 @@ val Context.desiredDialogSize: IntArray
         return dialogSize
     }
 
-// This check returns false in emulators
-val Context.hasSoftNavigationBar: Boolean
```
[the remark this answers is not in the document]
amitavsonims: I found the check faulty on Pixel 5 emulator as well as physical device. As we get 0 height of navbar in case it is missing, removing this check should be beneficial.

fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
amitavsonims: Both of these return the same values in most conditions. This changes helps in solving shifts(due to status bar) when device is rotated 90 ACW.
 I noticed that at 90 ACW, tooltip was getting shifted to the right by the height of the status bar.

fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
amitavsonims: Yes, I have attached the listener there, Line 160. So from here we dismiss the popupWindow and then it calls the dismissSideEffects.

fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
amitavsonims: We calculate the softNavBarOffset only at 270degrees when navbar is on the left. 
These were producing errors in my application, have tested this change there.

fluentui_transients/src/main/java/com/microsoft/fluentui/tooltip/Tooltip.kt
```diff
@@ -295,12 +302,12 @@ class Tooltip {
...
+        val rightEdge = ( startPosition + upArrowWidth + cornerRadius + margin - context.softNavBarOffsetX > displayWidth ) || (doesNotFitAboveOrBelow && anchorRect.left > rightSpace)
         val leftEdge = ( startPosition - cornerRadius - margin - context.softNavBarOffsetX < 0 ) || (doesNotFitAboveOrBelow && anchorRect.left < rightSpace)
```
aishwarya56, line 302: I believe this should be removed then?

amitavsonims: softNavBarOffset is only calculated at 270degree rotation, 0 otherwise. So this should be good.
image

aishwarya56 (approved):

PR #120, opened 2021-11-23
Support contentDescription in PersistentBottomSheet and support focus

aishwarya56:
Support contentDescription in PersistentBottomSheet and support focus changes in Drawer.
Now we can decide if we want to announce focus when expanding, if needed


nkhattar78 (approved):
Changes looks good.

nkhattar78 (approved):

PR #136, opened 2021-12-23
PersistentBottomSheet, when is magnified using Magnification accessibility tool, on touching of it beyond original boundary dismissing it.

yharsh:
### Platforms Impacted
- [ ] Android

### Description of changes

(a summary of the changes made, often organized by file)

Instead of relying on the getGlobalVisibleRect which returns original/unmagnified coordinates, attached a click listener on the outline which takes care of dismissing the bottom sheet.

### Verification

After fix, able to click on the menu items in persistent bottom sheet, able to dismiss it by clicking out the menu view.

### Pull request checklist

This PR has considered:


fluentui_drawer/src/main/java/com/microsoft/fluentui/persistentbottomsheet/PersistentBottomSheet.kt
```diff
@@ -279,6 +257,18 @@ class PersistentBottomSheet @JvmOverloads constructor(context: Context, attrs: A
         }
     }
 
+    private fun setClickListenerForDismissal() {
+        persistentSheetBinding.persistentBottomSheetOutlined.setOnClickListener {
+            if (persistentSheetBehavior.state == BottomSheetBehavior.STATE_EXPANDED) {
+                persistentSheetBehavior.state = BottomSheetBehavior.STATE_COLLAPSED
+            }
+            persistentSheetBinding.persistentBottomSheetOutlined.isClickable = false
+            if(Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
aishwarya56, line 266: Why are we doing this only for Android above O?
this can be moved to XML, is there is a limitation on API level

fluentui_drawer/src/main/java/com/microsoft/fluentui/persistentbottomsheet/PersistentBottomSheet.kt
```diff
@@ -257,19 +233,21 @@ class PersistentBottomSheet @JvmOverloads constructor(context: Context, attrs: A
 
     fun expand(focusDrawerHandle: Boolean = true) {
         persistentSheetBehavior.state = BottomSheetBehavior.STATE_EXPANDED
+        setClickListenerForDismissal()
```
aishwarya56, line 236: Cant we have this somewhere in init block?
That way. we do not need to set it individually for expand/show
Also, i feel when the sheet is gone then clicklistener wont work. so, this would work for us and we will not have to call it everytime the sheet is expanded.

yharsh: Sheet outline is there even after collapse but it's transparent and so when items are collapsed then we want to remove this listener which is done by setting clickable/focusable as false.

fluentui_drawer/src/main/java/com/microsoft/fluentui/persistentbottomsheet/PersistentBottomSheet.kt
yharsh: When sheet is expanded then we want to absorb events, while it's collapse we don't want to. And as outlined is there, even after collapse, we set this to unfocusable in order to not capture events and pass them to below view. Setting focusable=false is required in order to not absorb events for >=O.

defihook (commented):
graphics.Rect

PR #146, opened 2022-02-13
Calendar

mishramayank1:
1 fluentui_calendar : Moved the declare-styleable attributes to fluentui_core attrs.xml & enum attributes to fluentui_calendar attrs.xml to reuse at the module level. Attributes are also prefixed with 'fluentui_' to avoid conflict with other UI library attributes


fluentui_core/src/main/res/values/attrs.xml
```diff
@@ -34,4 +34,37 @@
     <!--Transparent Foregrounds-->
     <attr name="fluentuiForegroundOnPrimary80Color" format="reference|color"/>
     <attr name="fluentuiForegroundOnPrimary70Color" format="reference|color"/>
+
+    <!--Attributes from Modules-->
```
aishwarya56, line 38: Do we need this?

aishwarya56 (approved):

mishramayank1:
Merging fluentui_calendar

fluentui_core/src/main/res/values/attrs.xml
mishramayank1: These are existing attributes

PR #249, opened 2022-09-08
Shwshar1508 searchbar action view fix

shwshar1508:
### Platforms Impacted
- [ ] Android

### Description of changes

These change fix the bug in AppBarLayout, 
SearchBar -> ActionView type -> tap on search icon and start typing the text, cancel button appears at the center of the search bar instead of being at the right most ie. at the end of the SearchBar
We changed the root Layout from FrameLayout to RelativeLayout to fix this issue

### Verification

These changes are tested on Emulator Nexus 9 tablet, Samsung tablet and Pixel 4 emulator

Before screenshot:

Screenshot 2022-09-08 at 2 14 10 PM

After Screenshot:

Screenshot 2022-09-08 at 2 09 45 PM

This bug is more evident for Tablet devices

### DIFF
--- a/fluentui_topappbars/src/main/java/com/microsoft/fluentui/search/Searchbar.kt
+++ b/fluentui_topappbars/src/main/java/com/microsoft/fluentui/search/Searchbar.kt
@@ -18,6 +18,7 @@ import android.widget.FrameLayout
 import android.widget.ImageButton
 import android.widget.ImageView
 import android.widget.LinearLayout
+import android.widget.RelativeLayout
 import com.microsoft.fluentui.topappbars.R
 import com.microsoft.fluentui.appbarlayout.AppBarLayout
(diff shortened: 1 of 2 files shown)


fluentui_topappbars/src/main/java/com/microsoft/fluentui/search/Searchbar.kt
```diff
@@ -307,7 +308,7 @@ open class Searchbar : TemplateView, SearchView.OnQueryTextListener {
         )
 
         // Search view container
-        val searchViewContainerLayoutParams = searchViewContainer?.layoutParams as? FrameLayout.LayoutParams
+        val searchViewContainerLayoutParams = searchViewContainer?.layoutParams as? RelativeLayout.LayoutParams
```
mishramayank1, line 311: Please update the same at other places as well.

mishramayank1 (approved):

PR #289, opened 2022-11-29
ItemListContentBuilder

mishramayank1:
Created controls: TabItem, Header, Divider. Implemented ItemListContentBuilder api to create horizontal list, vertical grid, vertical list.

### DIFF
--- a/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
+++ b/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
@@ -4,6 +4,7 @@ import androidx.compose.animation.*
 import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.focusable
 import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.*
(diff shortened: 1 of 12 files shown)


fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
```diff
@@ -620,10 +621,13 @@ object ListItem {
...
             )
             val horizontalPadding = getListItemTokens().padding(Medium)
-            val leadPadding = if(leadingAccessoryView == null){
+            val leadPadding = if (leadingAccessoryView == null) {
                 PaddingValues(start = horizontalPadding, end = horizontalPadding)
-            }else {
-                PaddingValues(start = getListItemTokens().padding(size = GlobalTokens.SpacingTokens.None), end = horizontalPadding)
+            } else {
+                PaddingValues(
+                    start = getListItemTokens().padding(size = GlobalTokens.SpacingTokens.None),
```
NamanPandey, line 628: Should we keep ListItemInfo in the token functions as discussed for all controls?

NamanPandey (approved):

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
mishramayank1: Yes, Let's take it as separate PR for all the remaining controls.

PR #365, opened 2023-03-03
Set up localization workflow with Azure Pipelines

praveen970:


.github/workflows/localization.yml
```diff
@@ -14,11 +14,40 @@ jobs:
...
-        branch: tdbuild
-        labels: Automated PR
-        branch-suffix: timestamp
+      env:
+        GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+      run: |
+        set -e
+        echo "Start."
+        # Configure git and Push updates
+        git config --global user.email pyeruva@microsoft.com
```
vishnu2410, line 23: Is it possible to use some variable that gets populated at the run time or any other generic email?

PR #381, opened 2023-03-29
V2Persona List UI Tests

praveen970:
### Platforms Impacted
- [x] Android

### Description of changes

Adding UI Tests for V2PersonaLists


FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2PersonaListActivity.kt
```diff
@@ -236,6 +238,7 @@ class V2PersonaListActivity : DemoActivity() {
         val coroutineScope: CoroutineScope = rememberCoroutineScope()
         Scaffold(scaffoldState = scaffoldState) {
```
mishramayank1, line 239: Remove the Scaffold as this affect the internal element value. Use Toast to show onClick action

praveen970: Removed the scaffold usages in other PR

PR #436, opened 2023-07-03
modifying texticons to composables in listitem

praveen970:
### Problem 
Adding a badge on primary text in list item
### Root cause 
primary and secondary subText views are not composable so user cannot add any view.
### Fix
Convert the existing texticons view to composable so that any composable views can be added

### Validations

### Screenshots

image | image 

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
+++ b/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
@@ -77,12 +77,14 @@ object ListItem {
             BorderType.Top -> drawLine(
                 borderColor, Offset(0f, 0f), Offset(size.width, 0f), borderSize * density
             )
+
             BorderType.Bottom -> drawLine(
                 borderColor,
                 Offset(borderInset, size.height),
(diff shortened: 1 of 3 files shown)


fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
```diff
@@ -202,10 +206,10 @@ object ListItem {
         subTextMaxLines: Int = 1,
         secondarySubTextMaxLines: Int = 1,
         onClick: (() -> Unit)? = null,
-        primaryTextLeadingIcons: TextIcons? = null,
-        primaryTextTrailingIcons: TextIcons? = null,
-        secondarySubTextLeadingIcons: TextIcons? = null,
-        secondarySubTextTailingIcons: TextIcons? = null,
+        primaryTextLeadingView: (@Composable () -> Unit)? = null,
```
mishramayank1, line 209: As this is going to be breaking change, hence lets change suffix from "View" to "Content".  As "View" can mislead

mishramayank1 (approved):

PR #458, opened 2023-07-27
Added cornerRadius token in BadgeToken

mishramayank1:
### Problem 
Corner radius could not configure via token.
### Root cause 
Corner radius was hardcoded & hence token was not available in BadgeToken.
### Fix
Added cornerRadius in controlToken for badge to configure corner radius for badge.

### Validations
manual validation

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/BadgeTokens.kt
+++ b/fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/BadgeTokens.kt
@@ -8,6 +8,7 @@ import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.SolidColor
 import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import com.microsoft.fluentui.theme.FluentTheme
 import com.microsoft.fluentui.theme.token.ControlInfo
@@ -52,6 +53,7 @@ open class BadgeTokens : IControlToken, Parcelable {
                     badgeInfo
                 ).width
             )
+
             BadgeType.List -> PaddingValues(
                 start = FluentGlobalTokens.size(FluentGlobalTokens.SizeTokens.Size80) + borderStroke(
                     badgeInfo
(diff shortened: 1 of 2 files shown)


fluentui_notification/src/main/java/com/microsoft/fluentui/tokenized/notification/Badge.kt
```diff
@@ -72,15 +71,15 @@ fun Badge(
         val textColor = token.textColor(badgeInfo = badgeInfo)
         val typography = token.typography(badgeInfo = badgeInfo)
         val paddingValues = token.padding(badgeInfo = badgeInfo)
+        val shape = RoundedCornerShape(token.cornerRadius(badgeInfo = badgeInfo))
 
         Row(
             modifier
                 .requiredHeight(if (badgeType == BadgeType.Character) 20.dp else 27.dp)
```
praveen970, line 78: can we also token height as well.

fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/BadgeTokens.kt
```diff
@@ -72,4 +74,7 @@ open class BadgeTokens : IControlToken, Parcelable {
             FluentTheme.aliasTokens.neutralStrokeColor[FluentAliasTokens.NeutralStrokeColorTokens.StrokeFocus1].value()
         )
     }
+
+    @Composable
+    open fun cornerRadius(badgeInfo: BadgeInfo): Dp = 100.dp
```
NamanPandey, line 79: Can we use the CornerRadius Token for circle here instead?

fluentui_notification/src/main/java/com/microsoft/fluentui/tokenized/notification/Badge.kt
mishramayank1: Constraint for height is provided from design. Hence in Figma, there is no token for height as well.

fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/BadgeTokens.kt
mishramayank1: Ideally, we could but for now using the values which is mentioned in Figma. I have added the comment asking same question.

praveen970 (approved):

PR #492, opened 2023-08-29
CardNudge: RTL issue for swiping left/right

Anubhvv:
### Problem 
Swiping animation was not following the direction of swipe action (left or right).
The background "hide" text was not aligned correctly in RTL
### Root cause 
ReverseDirection was defaulting to "false" in swipeable.
Also, positive/ negative value for "text".
### Validations

### Screenshots

| Before                                                        |  After                                                      |
|  | |

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_notification/src/main/java/com/microsoft/fluentui/tokenized/notification/CardNudge.kt
+++ b/fluentui_notification/src/main/java/com/microsoft/fluentui/tokenized/notification/CardNudge.kt
@@ -134,15 +134,9 @@ fun CardNudge(
         ) {
             LaunchedEffect(state.offset.value) {
                 if (state.offset.value > 0.1F) {
-                    when {
-                        isRtl -> metadata.rightSwipeGesture?.invoke(-state.offset.value / maxWidth) //negative value for RTL for aligning the "Hide" Text appropriately
-                        else -> metadata.rightSwipeGesture?.invoke(state.offset.value / maxWidth)
-                    }
+                    metadata.rightSwipeGesture?.invoke(state.offset.value / maxWidth)
                 } else if (state.offset.value < -0.1F) {
-                    when {
-                        isRtl -> metadata.leftSwipeGesture?.invoke(-state.offset.value / maxWidth)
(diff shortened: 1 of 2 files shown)


fluentui_notification/src/main/java/com/microsoft/fluentui/tokenized/notification/CardNudge.kt
```diff
@@ -130,9 +134,15 @@ fun CardNudge(
         ) {
             LaunchedEffect(state.offset.value) {
                 if (state.offset.value > 0.1F) {
-                    metadata.rightSwipeGesture?.invoke(state.offset.value / maxWidth)
+                    when {
+                        isRtl -> metadata.rightSwipeGesture?.invoke(-state.offset.value / maxWidth) //negative value for RTL for aligning the "Hide" Text appropriately
+                        else -> metadata.rightSwipeGesture?.invoke(state.offset.value / maxWidth)
```
mishramayank1, line 139: Revert this change. It is consumer responsibility to process it as per layout direction.

fluentui_notification/src/main/java/com/microsoft/fluentui/tokenized/notification/CardNudge.kt
```diff
@@ -130,9 +134,15 @@ fun CardNudge(
...
+                    }
                 } else if (state.offset.value < -0.1F) {
-                    metadata.leftSwipeGesture?.invoke(state.offset.value / maxWidth)
+                    when {
```
NamanPandey, line 142: same as above

fluentui_notification/src/main/java/com/microsoft/fluentui/tokenized/notification/CardNudge.kt
Anubhvv: Done,
also  fixed this: (Now it shows "Left Swiped: some value"
Screenshot_2023-08-30-15-23-13-45_138c6ebfedc8e8a8cd1938476bab217a

mishramayank1 (approved):

PR #525, opened 2023-10-06
ListItem Section Header's Bottom Border Not visible fix

Anubhvv:
### Problem 
Bottom Border of sectionHeaders is NOT visible
### Root cause 
The SectionHeader is enompassed within a box, and borders are drawn in this box, this box was getting overlapped slightly by the listitem below it, hence Bottom border wasn't visible
### Fix
Instead of drawing the borders in Box, we are drawing it inside Column (which is nested inside Box).

### Validations

### Screenshots

|image|Screenshot_2023-10-06-17-07-30-98_138c6ebfedc8e8a8cd1938476bab217a|
|Screenshot_2023-10-06-17-08-16-45_ad3b258aa0cd3ec091b47fe7b71fec5b|Screenshot_2023-10-06-17-08-27-46_138c6ebfedc8e8a8cd1938476bab217a|

### Pull request checklist

This PR has considered:
- [x] Documentation and demo app examples

### DIFF
--- a/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
+++ b/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
@@ -1020,7 +1020,6 @@ object ListItem {
                 .fillMaxWidth()
                 .heightIn(min = cellHeight)
                 .background(backgroundColor)
-                .borderModifier(border, borderColor, borderSize, borderInsetToPx)
                 .focusable(false)
         ) {
             Row(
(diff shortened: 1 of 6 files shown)


mishramayank1 (approved):

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
```diff
@@ -1020,14 +1020,14 @@ object ListItem {
                 .fillMaxWidth()
```
praveen970, line 1020: Is this bix needed, it is same as the Row. Maybe we can remove the Box here

praveen970 (approved):

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/listitem/ListItem.kt
Anubhvv: Tested it.

PR #528, opened 2023-10-11
BasicInput Activity Accessibility Related Improvements

Anubhvv:
### Problem 

1. "Button" heading is not descriptive enough
2. "Button" is not defined and announced as heading, might be confusing for some people relying on accessibility features.
3. Talback is announcing "Disabled Button Button" for "Disableb Button", because of it's text, so giving better text to the buttons

### Fix

1. "Different views of button" is used as heading.
2. textView, setting accessibilityHeading = true.
3. Changed the text for buttons, appended example in text to differentiate

### Validations

### Screenshots

|image|Screenshot_2023-10-11-12-23-59-63_138c6ebfedc8e8a8cd1938476bab217a|

### Pull request checklist

This PR has considered:

### DIFF
--- a/FluentUI.Demo/src/main/res/layout/activity_basic_inputs.xml
+++ b/FluentUI.Demo/src/main/res/layout/activity_basic_inputs.xml
@@ -17,6 +17,7 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:paddingBottom="@dimen/demo_headline_padding_bottom"
+        android:accessibilityHeading="true"
         android:text="Different views of button" />
 
     <View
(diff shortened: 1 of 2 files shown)


FluentUI.Demo/src/main/res/layout/activity_basic_inputs.xml
```diff
@@ -17,7 +17,8 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:paddingBottom="@dimen/demo_headline_padding_bottom"
-        android:text="@string/button" />
+        android:accessibilityHeading="true"
+        android:text="Different views of button" />
```
mishramayank1, line 21: Add "Different views of button" string in String.xml & refer it.

Anubhvv: Getting that translation error when adding in string

mishramayank1 (approved):

PR #530, opened 2023-10-11
fixing crash on screen rotation

praveen970:
### Problem 
focusRequester is causing crash on recomposition
### Root cause 
focusRequester should be used on events. here focusRequester is not used on events causing the crash on recomposition
### Fix
Removed the focusrequeter since it is not required as the work is done using the TextField

### Validations

### Screenshots

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/tokenized/peoplepicker/PeoplePicker.kt
+++ b/fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/tokenized/peoplepicker/PeoplePicker.kt
@@ -24,8 +24,6 @@ import androidx.compose.runtime.snapshots.SnapshotStateList
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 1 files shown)


fluentui_peoplepicker/src/main/java/com/microsoft/fluentui/tokenized/peoplepicker/PeoplePicker.kt
```diff
@@ -194,7 +194,6 @@ fun PeoplePicker(
                             )
                             Spacer(modifier = Modifier.width(chipHorizontalSpacing))
                         }
-                        focusRequester.requestFocus()
                     }
```
mishramayank1, line 197: as this call has not been used then remove the focusRequester code from modifier & import as well.

praveen970: My bad, somehow changes dint add to this PR. Resolved

mishramayank1 (approved):

PR #533, opened 2023-10-16
adding undefined object for tokenset

praveen970:
### Problem 
Cannot conditionally evalute if the tokenset returns any control token if non existent controltype is used
### Root cause 
ControlTokenset throws runtime exception if not control type is found causing crash.
### Fix
Created undefined object of IControlToken which will be returned if ControlType is not found

### Validations

### Screenshots

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/ControlTokens.kt
+++ b/fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/ControlTokens.kt
@@ -25,6 +25,8 @@ interface IControlTokens {
     }
 }
 
+class UndefinedControlType: IControlToken
+
 /**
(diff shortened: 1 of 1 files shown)


fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/ControlTokens.kt
```diff
@@ -25,6 +25,8 @@ interface IControlTokens {
     }
 }
 
+class UndefinedControlType: IControlToken
```
mishramayank1, line 28: rename to UndefinedControlToken. Change its type to object

mishramayank1 (approved):

PR #550, opened 2023-11-02
V1 Persistent Bottom Sheet Accessibility Bug: Talkback Focus goes in background when Expended

Anubhvv:
### Problem 
When PersitentBottomSheet is in EXPANDED state, and talkback focus is navigated, it also goes to background views that are hidden by the sheet.
### Root cause 
Important For Accessibility for background views not dynamically changing.
### Fix
Creating a member variable backgroundViews (list of Views) which can be set as the views that should not get focus when sheet is expanded. 
Expand(), Hide(), Collapse() etc functions take care of setting importantForAccessibiity attritbute for those views.

### Validations

### Screenshots

Before:

https://github.com/microsoft/fluentui-android/assets/68989156/37a915f6-24b3-4a40-b64f-cef5fe891ee7

After:

https://github.com/microsoft/fluentui-android/assets/68989156/496acdb5-fef4-410d-8223-1ef9668814a6

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_drawer/src/main/java/com/microsoft/fluentui/persistentbottomsheet/PersistentBottomSheet.kt
+++ b/fluentui_drawer/src/main/java/com/microsoft/fluentui/persistentbottomsheet/PersistentBottomSheet.kt
@@ -54,6 +54,7 @@ class PersistentBottomSheet @JvmOverloads constructor(context: Context, attrs: A
     private var isDrawerHandleVisible = true
     private var sheetContainer: PersistentBottomSheetContentViewProvider.SheetContainerInfo? = null
     private var focusDrawerHandleInAccessibility = true
+    var backgroundViews: List<View>? = null
 
 
     init {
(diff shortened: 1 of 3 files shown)


fluentui_drawer/src/main/java/com/microsoft/fluentui/persistentbottomsheet/PersistentBottomSheet.kt
```diff
@@ -249,6 +251,7 @@ class PersistentBottomSheet @JvmOverloads constructor(context: Context, attrs: A
         if(focusDrawerHandle) {
             persistentSheetBinding.sheetDrawerHandle.requestFocus()
         }
+        setImportantForAccessibility(backgroundViews!!, View.IMPORTANT_FOR_ACCESSIBILITY_YES)
```
praveen970, line 254: This might cause NPE. Can we add check instead

Anubhvv: Done

praveen970 (approved):

PR #600, opened 2024-01-25
Drawer position flag

praveen970:
### Problem 
Drawer recent clipping changes caused a breaking change. 
### Root cause 
Drawer popup properties are update to not clip to the screen window
### Fix
Created a flag to whether clip to the boundaries or create a relative position for drawer popup

### Validations
Manual testing

### Screenshots

When relative bounds is false
Screenshot_20240125-143327
When relative bounds is true
Screenshot_20240125-143335

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2DrawerActivity.kt
+++ b/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2DrawerActivity.kt
@@ -206,27 +206,6 @@ private fun CreateActivityUI() {
                     )
                 }
                 item {
-                    val relativeToParentAnchorText =
-                        stringResource(id = R.string.drawer_relative_to_parent_anchor)
-                    ListItem.Header(title = relativeToParentAnchorText, modifier = Modifier
-                        .toggleable(
-                            value = relativeToParentAnchor,
-                            role = Role.Switch,
-                            onValueChange = { relativeToParentAnchor = !relativeToParentAnchor }
-                        )
-                        .clearAndSetSemantics {
-                            this.contentDescription = relativeToParentAnchorText
-                        }, trailingAccessoryContent = {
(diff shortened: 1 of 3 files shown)


Anubhvv (approved):

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2DrawerActivity.kt
```diff
@@ -68,220 +80,315 @@ private fun CreateActivityUI() {
...
-                        listContent = false
-                    },
-                    trailingAccessoryContent = {
-                        RadioButton(
-                            onClick = {
-                                dynamicSizeContent = true
-                                nestedDrawerContent = false
-                                listContent = false
+                        .clearAndSetSemantics {
+                            this.contentDescription = "relative bounds"
```
mishramayank1, line 217: update text here as well & localize it

fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/drawer/Drawer.kt
```diff
@@ -966,6 +967,7 @@ fun Drawer(
     behaviorType: BehaviorType = BehaviorType.BOTTOM,
     drawerState: DrawerState = rememberDrawerState(),
     scrimVisible: Boolean = true,
+    placeRelativeToAnchor: Boolean = false,
     offset: IntOffset = IntOffset(0, 0),
```
mishramayank1, line 971: Let make offset nullable. For null value pass clipingEnable = true in PopupPropeties & use offset (0,0) bydefault. If offset provided then pass clipingEnable = false & use the same offset. This way changes would be non-breaking at both Officemobile & Loop.

mishramayank1 (approved):

PR #626, opened 2024-03-16
Compose UI dependency update

itsPronay:


defihook (commented):
seems like version upgraded

PR #641, opened 2024-04-19
adding keycorrction to not duplicate the offset keys

praveen970:
### Problem 
bottomsheet anchor map replaces statevalues when the keys are same in some scenarios
### Root cause 
Map used for anchors use Float as Key. These key values can be same in some scenarios
### Fix
Added a very small value for key that will prevent from key duplication while also not causing any issues wrt to dp

### Validations
Manual Testing

### Pull request checklist

This PR has considered:
- [ ] Automated Tests


rohanprasadofficial (approved):

mishramayank1:
Add value to Hidden state instead of Shown or Expanded state as this wont affect UX alignment anyway.

PR #659, opened 2024-06-13
bottom sheet width set to 60% in landscape mode and aligned center

ayanshakil:
bottomsheet width can now be set to 60 % in landscape mode and aligned to center of the screen.


fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/BottomSheetTokens.kt
```diff
@@ -46,4 +46,11 @@ open class BottomSheetTokens : IControlToken, Parcelable {
 
     @Composable
     open fun scrimOpacity(bottomSheetInfo: BottomSheetInfo): Float = 0.32F
+
+    @Composable
+    open fun maxLandscapeWidth (bottomSheetInfo: BottomSheetInfo): Float =
+        0.6F
```
praveen970, line 52: make this as 1F default or else it will be a breaking for existing partners

ayanshakil:
> @microsoft-github-policy-service agree

praveen970 (approved):

PR #667, opened 2024-06-28
v2 Tab Bar: Implementing automatic text adjustment to prevent overflow or multiline issues and removing padding on smaller devices.

Anubhvv:
### Problem 
The text was overflowing and being present in multiple lines for tab items in tab bar. 
Requirement is for text to remain in one line. 

There are 2 things we can do here:
For any screen width smaller than 360dp, the side padding (16dp on either sides) will reduce proportionately to accommodate text. 

 
2. We can reduce text size in accordance to fit in one line. What we do on Title bar currently.
### Root cause 

By design

### Fix

1.  we are detecting text overflow using the onTextLayout callback. This allows us to recursively reduce text size untill it fits in the avaibale one line width
2. For small devices have removed the 16.dp padding

### Validations

### Screenshots

| small phone                          | big phone                                   |
|image| image |

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
+++ b/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
@@ -21,12 +21,13 @@ import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
-import androidx.compose.ui.unit.sp
 import androidx.constraintlayout.compose.ChainStyle
 import androidx.constraintlayout.compose.ConstraintLayout
 import androidx.constraintlayout.compose.Dimension
 import com.microsoft.fluentui.theme.FluentTheme
 import com.microsoft.fluentui.theme.token.ControlTokens
+import com.microsoft.fluentui.theme.token.FluentAliasTokens
+import com.microsoft.fluentui.theme.token.FluentGlobalTokens
 import com.microsoft.fluentui.theme.token.FluentStyle
 import com.microsoft.fluentui.theme.token.Icon
 import com.microsoft.fluentui.theme.token.controlTokens.TabItemInfo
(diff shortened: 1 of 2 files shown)


fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
```diff
@@ -202,11 +208,22 @@ fun TabItem(
                 .then(widthModifier)
         ) {
             badgeWithIcon()
+            var fontSize = remember { mutableStateOf(14.sp) }
+            var textStyle by remember(textColor) { mutableStateOf(TextStyle(color = textColor, textAlign = TextAlign.Center, fontSize = fontSize.value)) }
+
             if (textAlignment == TabTextAlignment.VERTICAL) {
                 Spacer(modifier = Modifier.height(2.dp))
-                BasicText(
+                Text(
```
praveen970, line 216: Lets not use material related APIs. Any reason why Basic text is not used. basictext is compose API

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
```diff
@@ -202,11 +208,22 @@ fun TabItem(
                 .then(widthModifier)
         ) {
             badgeWithIcon()
+            var fontSize = remember { mutableStateOf(14.sp) }
```
praveen970, line 211: Why 14.sp for fontsize?

fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
```diff
@@ -59,7 +60,7 @@ fun TabBar(
         )
         Row(
             modifier = Modifier
-                .padding(horizontal = 16.dp)
+                .padding(horizontal = if(LocalConfiguration.current.screenWidthDp.dp > 360.dp )16.dp else 1.dp)       // For small phones we don't want the text overflowing, hence removing padding
```
praveen970, line 63: What if the device is small and the number of items are 2. There will be no padding and is this expected? Should we removed the padding only when the items are more?

Anubhvv: small phone
There's plenty of space in less number of items

praveen970 (approved):

PR #706, opened 2024-10-08
Added color scheme in horizontalListAdapter

Dhruv-Mishra:
### Problem 
Icons were not properly visible in PersistentBottomSheet

### Root cause 
When passing the items to HorizontalListAdapter, it was passed without the required color tint leading to default value which was different from the rest of the bottomsheet items 

### Fix
Added ability to pass a color tint as a ColorInt variable, and aligned the color with the rest of the bottomsheet elements

### Validations

##### Before fix in Light Mode: 
image

##### Before fix in Dark Mode: 
image

##### After fix in Light Mode: 
image

##### After fix in Dark Mode: 
image

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/PersistentBottomSheetActivity.kt
+++ b/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/PersistentBottomSheetActivity.kt
@@ -240,7 +240,7 @@ class PersistentBottomSheetActivity : DemoActivity(), SheetItem.OnClickListener,
                     ContextCompat.getColor(this, R.color.bottomsheet_horizontal_icon_tint),
                     disabled = true
                 )
-            ), 0, marginBetweenView
+            ), 0, marginBetweenView, drawerTint = ContextCompat.getColor(this, R.color.bottomsheet_horizontal_icon_tint).toInt()
         )
         horizontalListAdapter.mOnSheetItemClickListener = this
         persistentSheetContentBinding.sheetHorizontalItemList3.layoutManager =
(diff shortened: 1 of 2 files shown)


joyeeta26:
from the screenshot, icons still doesn't seem to be visible in dark mode

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/PersistentBottomSheetActivity.kt
```diff
@@ -240,7 +240,7 @@ class PersistentBottomSheetActivity : DemoActivity(), SheetItem.OnClickListener,
                     ContextCompat.getColor(this, R.color.bottomsheet_horizontal_icon_tint),
```
joyeeta26, line 240: from the screenshot, icons still doesn't seem to be visible in dark mode

Dhruv-Mishra: We are following the central Fluent Theming and colors here, the problem was the consistency of the icon, this one row, in specific, had been different from the rest of the icons. I am attaching more screenshots of the entire bottomDrawer for your reference.
If we want to change the color for these, then that would have to be done for the entire element, the aim for this PR was to make them consistent throughout the UI.

Dhruv-Mishra: I have updated the screenshots, please check now
The perceived lack of visibility may also be due to the fact that the icons that have been updated are set to disabled state, i.e, disabled for user interaction and thus are dimmer than the rest

Dhoot (approved):

PR #708, opened 2024-10-14
Allowed use of gradient in tabItem Indicator

Dhruv-Mishra:
### Problem 
TabItem Indicator in Tab Bar had fixed color drawn from text, modified it to accept custom colors and gradients

### Root cause 
Indicator color was directly assigned as text color 

### Fix
Added arguments to assign indicator color 

### Screen recording

| [beforeIndicator.webm](https://github.com/user-attachments/assets/8af539d0-b4d6-4466-ba67-4596c3f3f987) | [afterIndicator.webm](https://github.com/user-attachments/assets/a1797c95-2ae5-4a81-a510-a627611a4180) |

Color Scheme Used in the After Video: 
image

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/TabItemTokens.kt
+++ b/fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/TabItemTokens.kt
@@ -3,6 +3,7 @@ package com.microsoft.fluentui.theme.token.controlTokens
 import android.os.Parcelable
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.SolidColor
 import androidx.compose.ui.text.TextStyle
@@ -23,7 +24,8 @@ enum class TabTextAlignment {
 
 open class TabItemInfo(
     val tabTextAlignment: TabTextAlignment = TabTextAlignment.VERTICAL,
-    val fluentStyle: FluentStyle = FluentStyle.Neutral
+    val fluentStyle: FluentStyle = FluentStyle.Neutral,
+    val indicatorColor: StateBrush? = null
 ) : ControlInfo
 
 @Parcelize
@@ -170,5 +172,10 @@ open class TabItemTokens : IControlToken, Parcelable {
     @Composable
     open fun textTypography(tabItemInfo: TabItemInfo): TextStyle {
         return FluentTheme.aliasTokens.typography[FluentAliasTokens.TypographyTokens.Caption2]
+    }
+
+    @Composable
+    open fun indicatorColor(tabItemInfo: TabItemInfo): StateBrush? {
+        return tabItemInfo.indicatorColor
     }
 }
--- a/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
+++ b/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
@@ -2,6 +2,7 @@ package com.microsoft.fluentui.tokenized.tabItem
 
 import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.animateColorAsState
+import androidx.compose.animation.core.animateValueAsState
 import androidx.compose.animation.core.tween
 import androidx.compose.animation.expandHorizontally
 import androidx.compose.animation.fadeIn
@@ -22,6 +23,7 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.BasicText
 import androidx.compose.material.ripple.rememberRipple
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.State
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
@@ -29,6 +31,8 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.layout.Layout
 import androidx.compose.ui.layout.layoutId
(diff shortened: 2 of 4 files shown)


fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
```diff
@@ -76,6 +78,7 @@ fun TabBar(
                     onClick = tabData.onClick,
                     accessory = tabData.badge,
                     showIndicator = showIndicator,
+                    indicatorStyle = indicatorStyle,
```
joyeeta26, line 81: since we are already passing this separately, why have we included this in tabItemTokens?

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
```diff
@@ -274,14 +283,22 @@ fun TabItem(
                 )
                 {
                     Box(
-                        modifier = Modifier
+                        modifier = if(indicatorColor != null) {
+                            Modifier
                             .height(3.dp)
                             .width(indicatorWidth)
-                            .background(shape = RoundedCornerShape(indicatorCornerRadiusSize), color = textColor)
+                            .background(shape = RoundedCornerShape(indicatorCornerRadiusSize), brush = indicatorColor)
```
joyeeta26, line 290: why are we using brush and not color?

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
```diff
@@ -72,7 +78,7 @@ fun TabItem(
         tabItemTokens
             ?: FluentTheme.controlTokens.tokens[ControlTokens.ControlType.TabItemControlType] as TabItemTokens
 
-    val tabItemInfo = TabItemInfo(textAlignment, style)
```
joyeeta26, line 75: Also we need color for icon as well

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
Dhruv-Mishra: Brush is itself an Object containing all the information about gradient and colors used. As far as I am aware, it is not possible to use gradient as a color object neither is it possible to cast Brush to Color.

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
Dhruv-Mishra: This PR is for TabItem Indicator only, for Icon we will have to modify Icon class as well to support a Brush object, will add that in the next PR.

fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
Dhruv-Mishra: Will remove arguments for indicatorStyle and Text alignment in the upcoming PR and merge them into tabItemInfo to pass as a single argument instead

fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/TabItemTokens.kt
```diff
@@ -171,4 +173,9 @@ open class TabItemTokens : IControlToken, Parcelable {
     open fun textTypography(tabItemInfo: TabItemInfo): TextStyle {
         return FluentTheme.aliasTokens.typography[FluentAliasTokens.TypographyTokens.Caption2]
     }
+
+    @Composable
+    open fun indicatorColor(tabItemInfo: TabItemInfo): StateBrush? {
```
praveen970, line 178: return default fluent values here. In this case the values are same as the text color

fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/TabItemTokens.kt
```diff
@@ -23,7 +24,8 @@ enum class TabTextAlignment {
 
 open class TabItemInfo(
     val tabTextAlignment: TabTextAlignment = TabTextAlignment.VERTICAL,
-    val fluentStyle: FluentStyle = FluentStyle.Neutral
+    val fluentStyle: FluentStyle = FluentStyle.Neutral,
+    val indicatorColor: StateBrush? = null
```
praveen970, line 28: this is not required here.

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
```diff
@@ -59,6 +64,7 @@ fun TabItem(
     modifier: Modifier = Modifier,
     style: FluentStyle = FluentStyle.Neutral,
     textAlignment: TabTextAlignment = TabTextAlignment.VERTICAL,
+    indicatorStyle: StateBrush? = null,
```
praveen970, line 67: This is not necessary. You can take values directly from tokens

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
```diff
@@ -274,14 +283,22 @@ fun TabItem(
                 )
                 {
                     Box(
-                        modifier = Modifier
+                        modifier = if(indicatorColor != null) {
```
praveen970, line 286: this condition wont be needed. Just use the exisitng implementation and add the color form the indicator color token

fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
```diff
@@ -45,6 +46,7 @@ fun TabBar(
     selectedIndex: Int = 0,
     tabTextAlignment: TabTextAlignment = TabTextAlignment.VERTICAL,
     showIndicator: Boolean = false,
+    indicatorStyle: StateBrush? = null,
```
praveen970, line 49: This will be removed

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
Dhruv-Mishra: This is needed because of function overloading of the Modifier.background function. There are two different implementations of the background function being called.
If the background is a Color object, then we use the implementation which takes Color as an argument, if the background is a Brush object, then we use the implementation which takes Brush as an argument.

fluentui_core/src/main/java/com/microsoft/fluentui/theme/token/controlTokens/TabItemTokens.kt
Dhruv-Mishra: If the indicatorColor returns null then the indicator color will be set to textColor, the logic for that has been added when setting the indicator color

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
praveen970: Once token is used here then it will be Brush and not Color

PR #715, opened 2024-10-17
Scrim functionality fixes & other enhancements for v2 bottom sheet

joyeeta26:
### Problem 
This PR fixes multiple problems - 
1. Scrim was not getting visible and was not working as expected.
2. no way was present to perform some action after dismiss
3. Preventing dismissal on scrim click was not available.
4. Scrim functionality was not centralized in the drawer module. 
5. expand button was unusable in the demo bottom sheet activity
6. Peek height option in demo app was always disabled

### Root cause 
1. Fraction calculation was not happening correctly, hence scrim was not getting visible.
2. onDismiss callback was not present in bottom sheet arguments
3. preventing dismissal on scrim click option was also not available
4. Bottom sheet and Drawer were having their own implementations for Scrim
5. This was because the expand button was getting activated only when the drawer is visible, and in that case, due to scrim the button was getting unusable.
6. Peek height was always disabled because condition for enabling was when bottom sheet is visible and in that scenario due to scrim, the button becomes unusable

### Fix
1. Fixed the fraction calculation, now scrim is visible.
2. Added onDismiss callback
3. Added preventDismissalOnScrimClick option
4. Centralized Scrim across drawer module.
5. Added the expand button to be always visible and removed the hide button as it was not required.
6. Updated the condition so that the peek height option remains visible when the bottom sheet is hidden

### Validations

After

bottom sheet

### DIFF
--- a/fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/bottomsheet/BottomSheet.kt
+++ b/fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/bottomsheet/BottomSheet.kt
@@ -242,7 +242,7 @@ fun BottomSheet(
     stickyThresholdUpward: Float = 56f,
     stickyThresholdDownward: Float = 56f,
     bottomSheetTokens: BottomSheetTokens? = null,
-    onScrimClick: () -> Unit = {},
+    onDismiss: () -> Unit = {},
     content: @Composable () -> Unit
 ) {
     val themeID =
@@ -318,7 +318,7 @@ fun BottomSheet(
                     }
                 },
                 open = sheetState.isVisible,
-                onScrimClick = onScrimClick,
+                onScrimClick = onDismiss,
                 preventDismissalOnScrimClick = preventDismissalOnScrimClick,
                 tag = BOTTOMSHEET_SCRIM_TAG
             )
@@ -391,6 +391,7 @@ fun BottomSheet(
                                 if (sheetState.confirmStateChange(BottomSheetValue.Hidden)) {
                                     scope.launch { sheetState.hide() }
                                 }
+                                onDismiss()
                                 true
                             }
                         }
@@ -438,6 +439,7 @@ fun BottomSheet(
                                         if (!sheetState.isVisible) {
                                             if (enableSwipeDismiss) {
                                                 scope.launch { sheetState.hide() }
+                                                onDismiss()
                                             } else {
                                                 scope.launch { sheetState.show() }
                                             }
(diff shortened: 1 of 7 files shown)


fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/bottomsheet/BottomSheet.kt
```diff
@@ -233,9 +238,11 @@ fun BottomSheet(
     showHandle: Boolean = true,
     slideOver: Boolean = true,
     enableSwipeDismiss: Boolean = false,
+    preventDismissalOnScrimClick: Boolean = false,
     stickyThresholdUpward: Float = 56f,
     stickyThresholdDownward: Float = 56f,
     bottomSheetTokens: BottomSheetTokens? = null,
+    onScrimClick: () -> Unit = {},
```
Anubhvv, line 245: BottomDrawer takes two params, onDismiss and onScrimClick, these two differ on when they run (post closing of drawer or pre), can we do the same thing here, so all scanerios are covered?

joyeeta26: BottomDrawer doesn't expose onDismiss

Dhoot (approved):

Anubhvv (approved):

Anubhvv (approved):

fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/bottomsheet/BottomSheet.kt
```diff
@@ -233,9 +238,11 @@ fun BottomSheet(
     showHandle: Boolean = true,
     slideOver: Boolean = true,
     enableSwipeDismiss: Boolean = false,
+    preventDismissalOnScrimClick: Boolean = false,
     stickyThresholdUpward: Float = 56f,
     stickyThresholdDownward: Float = 56f,
     bottomSheetTokens: BottomSheetTokens? = null,
+    onDismiss: () -> Unit = {},
```
Anubhvv, line 245: Add param description above the function

fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/bottomsheet/BottomSheet.kt
```diff
@@ -233,9 +238,11 @@ fun BottomSheet(
     showHandle: Boolean = true,
     slideOver: Boolean = true,
     enableSwipeDismiss: Boolean = false,
+    preventDismissalOnScrimClick: Boolean = false,
```
Anubhvv, line 241: same, update param description

fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/bottomsheet/BottomSheet.kt
joyeeta26: added

fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/bottomsheet/BottomSheet.kt
joyeeta26: added

Anubhvv (approved):

Anubhvv (approved):

PR #723, opened 2024-10-29
Fix alignment issue in list content builder & peek height option

joyeeta26:
### Problem 
1. Alignment issue in List content builder
2. Peek height option not usable in demo app bottom sheet

### Root cause 
1. The issue was happening specifically when maxItemsInRow > itemsInRow and equidistant= false, because width ratio was being calculated as 1/itemsInRow.
2. Because peek height option was getting activated when bottom sheet gets activated, hence rendering it unusable 

### Fix
1. For euidistant false, the width ratio should be calculated as 1/maxItemsInRow.
2. Updated the condition so that peek height option is visible when bottom sheet is hidden

### Validations
for maxItemsInRow > size
image

For maxItemsInRow = size
image

For maxItemsInRow < size
image

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2BottomSheetActivity.kt
+++ b/FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2BottomSheetActivity.kt
@@ -429,14 +429,14 @@ private fun CreateActivityUI() {
                     style = ButtonStyle.Button,
                     size = ButtonSize.Medium,
                     text = "+ 8 dp",
-                    enabled = !hidden,
+                    enabled = hidden,
                     onClick = { peekHeightState += 8.dp })
 
                 Button(
                     style = ButtonStyle.Button,
                     size = ButtonSize.Medium,
                     text = "- 8 dp",
-                    enabled = !hidden && (peekHeightState > 0.dp),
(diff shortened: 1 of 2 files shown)


FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2BottomSheetActivity.kt
```diff
@@ -429,14 +429,14 @@ private fun CreateActivityUI() {
                     style = ButtonStyle.Button,
                     size = ButtonSize.Medium,
                     text = "+ 8 dp",
-                    enabled = !hidden,
+                    enabled = hidden,
```
Anubhvv, line 432: not having context regarding this, can you tell me what is the reason for this change?

joyeeta26: so currently in demo app, peek height option is only shown when drawer activates
image
and once the drawer activates, the option being unusable as it's hidden behind the scrim
so here we are making the change so that peek height option is enabled when the drawer is not activated so that we can configure the peek height before activating the drawer

Anubhvv (approved):

Dhoot (approved):

PR #730, opened 2024-11-19
Release Version 0.3.3

Dhruv-Mishra:
Release Version 0.3.3:
What's new

Fluent v2:
Fix/enhancement

	1. Fixed Background Content Not Being Interactable even with scrim not visible in V2 BottomSheet
	2. Updated Default behaviour of Scrim to be not visible in V2 BottomSheet 

Note: Fluent versions 0.2.x will support compose version 1.5.1 (BOM 2023.09.00). Fluent version 0.3.x will support compose foundation version 1.6.0.


FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/Demos.kt
```diff
@@ -161,25 +161,25 @@ val V1DEMO = arrayListOf(
 )
 
 val V2DEMO = arrayListOf(
-    Demo(V2APP_BAR_LAYOUT, V2AppBarActivity::class, Badge.Modified),
```
joyeeta26, line 164: I think we can keep the badge for someting, because there were a lot of changes in the previous release

joyeeta26 (approved):

Dhoot (approved):

PR #744, opened 2024-12-02
Fixes for adding predictive back gestures in Modal Popup

Dhruv-Mishra:
### Problem 
When predictive back gestures were enabled for the components using Modal Popup (Bottom Drawer in our case), the back button press was not intercepted and thus the associated logic not handled.

### Root cause 
When  android:enableOnBackInvokedCallback is set to true for the app activity, Android system no longer sends the KEYCODE_BACK signal for back button presses. Using OnBackPressed dispatcher did not work with modal view either since it's view was not attached to the main activity.

### Fix
Updated the code for Modal Popup to allow callbacks registered in the main activity to apply to the modal view as well.

### Other Components
While this problem is also present in other drawers, they unfortunately do not use modal popup views. Will raise other PRs to fix them as well. This one should fix back behavior in Android 15 and above for all components relying on ModalPopups without modifying the API for the same. This will also allow for custom back behavior implementation,

### Screenshots
|  beforePredictive |  afterfix |

### Pull request checklist

This PR has considered:
- [ ] Automated Tests


fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -56,53 +67,70 @@ import java.util.UUID
  */
 @Composable
 fun ModalPopup(
-    onDismissRequest: () -> Unit,
+    onDismissRequest:(() -> Unit)? = null,
```
joyeeta26, line 70: why are we making this optional?

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2BottomDrawerActivity.kt
```diff
@@ -49,11 +50,17 @@ class V2BottomDrawerActivity : V2DemoActivity() {
...
+        override fun handleOnBackPressed() {
+            finish()
+        }
 
+    }
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setActivityContent {
             CreateActivityUI()
+            LocalOnBackPressedDispatcherOwner.current?.onBackPressedDispatcher?.addCallback(this, onBackCallback)
```
joyeeta26, line 63: can we add comment for this?

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -56,53 +67,70 @@ import java.util.UUID
...
-                            )
-                            .imePadding()
-                    ) {
-                        currentContent()
-                    }
+            setCustomContent(parentComposition) {
+                Box(
+                    Modifier
+                        .semantics { this.popup() }
+
```
joyeeta26, line 93: can we add the comments back?

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -56,53 +67,70 @@ import java.util.UUID
...
+                        .onSizeChanged {
+                            popupContentSize = it
+                        }
+
```
joyeeta26, line 97: here also

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -56,53 +67,70 @@ import java.util.UUID
...
             modalWindow.disposeComposition()
             modalWindow.dismiss()
         }
     }
+
+    Layout(
+        content = {},
+        modifier = Modifier
+            .onGloballyPositioned { childCoordinates ->
+                val parentCoordinates = childCoordinates.parentLayoutCoordinates!!
```
joyeeta26, line 127: can we avoid using !! operator if possible?

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -122,18 +150,32 @@ fun convertWindowInsetsCompatTypeToWindowInsets(windowInsetsCompatType: Int): Wi
 
 /** Custom compose view for [BottomDrawer] */
 private class ModalWindow(
-    private var onDismissRequest: () -> Unit,
+    private var onDismissRequest: (() -> Unit)? = null,
```
joyeeta26, line 153: why are we making this optional?

fluentui_drawer/build.gradle
```diff
@@ -52,6 +52,7 @@ dependencies {
     implementation "androidx.appcompat:appcompat:$appCompatVersion"
     implementation "androidx.recyclerview:recyclerview:$recyclerViewVersion"
     implementation "com.google.android.material:material:$materialVersion"
+    implementation 'androidx.activity:activity-compose:1.9.3'
```
joyeeta26, line 55: we have composeActivityVersion constant present in repo, can we use that?

fluentui_drawer/src/main/java/com/microsoft/fluentui/tokenized/drawer/Drawer.kt
```diff
@@ -497,9 +499,12 @@ fun BottomDrawer(
         val behaviorType =
             if (slideOver) BehaviorType.BOTTOM_SLIDE_OVER else BehaviorType.BOTTOM
         val drawerInfo = DrawerInfo(type = behaviorType)
+        BackHandler { //TODO: Add pull down animation with predictive back
```
joyeeta26, line 502: will this be added in follow up PR?

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -183,75 +302,48 @@ private class ModalWindow(
...
+        params: ViewGroup.LayoutParams
+    )
+}
 
-    fun dismiss() {
-        this.setViewTreeLifecycleOwner(null)
-        setViewTreeSavedStateRegistryOwner(null)
-        composeView.viewTreeObserver.removeOnGlobalLayoutListener(this)
-        windowManager.removeViewImmediate(this)
+private open class PopupLayoutHelperImpl : PopupLayoutHelper {
```
joyeeta26, line 319: formatting

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2DrawerActivity.kt
```diff
@@ -1,6 +1,9 @@
 package com.microsoft.fluentuidemo.demos
 
 import android.os.Bundle
```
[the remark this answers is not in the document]
Dhruv-Mishra: As stated in the PR description, this PR only involves the changes for bottom drawer. Will raise a separate PR for drawers.

FluentUI.Demo/src/main/java/com/microsoft/fluentuidemo/demos/V2BottomDrawerActivity.kt
Dhruv-Mishra: Will add this

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: It is no longer required, the default back press behavior will navigate back to the parent activity now

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: Will add back

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: Same as above

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: It is formatted, the popup layout helper methods are defined outside the class

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: Will add back

fluentui_drawer/build.gradle
Dhruv-Mishra: Bumped up the compose activity version and added the variable here

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: Updated

fluentui_drawer/build.gradle
joyeeta26: just double check, if bumping the version will cause any issues with partner teams

Dhruv-Mishra: Will reach out to them

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
joyeeta26: just to understand this more, this onDismissRequestCallback is for any action required to be perfomed on dismiss right?
Like even if it's navigating back to the parent activity, there could be some action required to be perfomed on dismiss?

FluentUI.Demo/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,7 @@
 
     <application
         android:allowBackup="true"
+        android:enableOnBackInvokedCallback="true"
```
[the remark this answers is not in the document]
joyeeta26: since we are now setting this as false for now, we can remove this only, until we decide to add it and set it as true

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -122,18 +152,32 @@ fun convertWindowInsetsCompatTypeToWindowInsets(windowInsetsCompatType: Int): Wi
 
 /** Custom compose view for [BottomDrawer] */
 private class ModalWindow(
-    private var onDismissRequest: () -> Unit,
+    private var onDismissRequest: (() -> Unit)? = null,
+    private var properties: PopupProperties,
     private val composeView: View,
+    density: Density,
     saveId: UUID,
+    private val popupLayoutHelper: PopupLayoutHelper = if (Build.VERSION.SDK_INT >= 29) {
```
joyeeta26, line 160: can you add a comment explaining, why are are using 2 different versions and what's different in these versions

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -183,75 +304,48 @@ private class ModalWindow(
...
-        this.content = content
-        shouldCreateCompositionOnAttachedToWindow = true
     }
+}
 
-    fun show() {
-        windowManager.addView(this, params)
-    }
+@VisibleForTesting
+internal interface PopupLayoutHelper {
```
joyeeta26, line 311: can we move this helper interfaces and implementation to some util file ?

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -183,75 +304,48 @@ private class ModalWindow(
...
-        parent?.let { setParentCompositionContext(it) }
-        this.content = content
-        shouldCreateCompositionOnAttachedToWindow = true
     }
+}
 
-    fun show() {
-        windowManager.addView(this, params)
-    }
+@VisibleForTesting
```
joyeeta26, line 310: why is this annotation needed?

FluentUI.Demo/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
Dhruv-Mishra: I will add this at the activity level instead of enabling it app wide

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: Yes, but the default behavior for any view is to close on back button press now.
So if we don't explicitly give a On Dismiss function, this will still work.

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: Will add this

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: Since the helper is only leveraged in Modal Window and Modal Window is only leveraged in BottomDrawer, I think we can include it here only.

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: I was trying with different Android versions, will remove this

fluentui_drawer/build.gradle
Dhruv-Mishra: It seems to be working fine with older activity versions as well. Will update this.

Dhruv-Mishra: No need for the Compose Activity version upgrade. Tested this down to API26 and it is performing as expected without the Compose Activity version upgrade.

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -56,53 +66,72 @@ import java.util.UUID
...
+            properties = properties,
+            layoutDirection = layoutDirection
+        )
         onDispose {
             modalWindow.disposeComposition()
             modalWindow.dismiss()
         }
     }
+
+    Layout(
```
joyeeta26, line 122: just for my understanding, what's the role of Layout here?

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -143,14 +186,91 @@ private class ModalWindow(
...
+
+    fun dismiss() {
+        this.setViewTreeLifecycleOwner(null)
+        setViewTreeSavedStateRegistryOwner(null)
+        windowManager.removeViewImmediate(this)
+    }
 
-    private val params: WindowManager.LayoutParams =
-        WindowManager.LayoutParams().apply {
+    override fun setLayoutDirection(layoutDirection: Int) {
```
joyeeta26, line 260: why are we overriding this with No-op

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
```diff
@@ -122,18 +151,32 @@ fun convertWindowInsetsCompatTypeToWindowInsets(windowInsetsCompatType: Int): Wi
 
 /** Custom compose view for [BottomDrawer] */
 private class ModalWindow(
-    private var onDismissRequest: () -> Unit,
+    private var onDismissRequest: (() -> Unit)? = null,
+    private var properties: PopupProperties,
     private val composeView: View,
+    density: Density,
     saveId: UUID,
+    private val popupLayoutHelper: PopupLayoutHelperImpl = if (Build.VERSION.SDK_INT >= 29) {
```
joyeeta26, line 159: can you add comment here or above the implementation below, why we need separate implementation for different versions?

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: For Android versions below API 29, it’s not necessary to explicitly exclude the entire screen from system gestures. However, starting with Android API 29, you need to explicitly specify if you want to exclude the entire screen area from gestures. The skeleton method is defined to keep consistency in the two objects.

fluentui_core/src/main/java/com/microsoft/fluentui/compose/ModalPopup.kt
Dhruv-Mishra: Added in the class definitions below

joyeeta26 (approved):

PR #747, opened 2024-12-08
[Accessibility] Tab Bar Talkback Fix

Dhruv-Mishra:
### Problem 
The selected tab was not announced separately for TabBar when talkback was turned on, possibly leading to accessibility challenges. This PR aims to fix that.

### Fix
Talkback now announces if a tab is currently active or inactive. This is configurable by the developer but there is a default implementation as well.

### Demonstration

| [beforeTabBarAccesibility.webm](https://github.com/user-attachments/assets/96c09eb3-a192-4ad9-ba2b-194fa05c0d47) | [tabBarAccesibilityFix.webm](https://github.com/user-attachments/assets/fabf6c3c-1d32-4dfb-8af3-eb985a5a04ed) |

### Pull request checklist

This PR has considered:
- [ ] Automated Tests

### DIFF
--- a/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
+++ b/fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
@@ -27,6 +27,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
+import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.draw.drawWithCache
@@ -37,6 +38,8 @@ import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.layout.Layout
 import androidx.compose.ui.layout.layoutId
 import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.semantics.invisibleToUser
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
@@ -53,6 +56,7 @@ import com.microsoft.fluentui.theme.token.controlTokens.TabItemInfo
 import com.microsoft.fluentui.theme.token.controlTokens.TabItemTokens
 import com.microsoft.fluentui.theme.token.controlTokens.TabTextAlignment
 
+@OptIn(ExperimentalComposeUiApi::class)
 @Composable
 fun TabItem(
     title: String,
@@ -117,7 +121,11 @@ fun TabItem(
     val iconContent: @Composable () -> Unit = {
         Icon(
             imageVector = icon,
-            modifier = Modifier.size(if (textAlignment == TabTextAlignment.NO_TEXT) 28.dp else 24.dp)
+            modifier = Modifier
+                        .semantics {
+                            invisibleToUser()
+                        }
+                        .size(if (textAlignment == TabTextAlignment.NO_TEXT) 28.dp else 24.dp)
                         .graphicsLayer(alpha = 0.99f)
                         .drawWithCache {
                             onDrawWithContent {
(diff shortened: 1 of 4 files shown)


fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
```diff
@@ -117,7 +121,11 @@ fun TabItem(
     val iconContent: @Composable () -> Unit = {
         Icon(
             imageVector = icon,
-            modifier = Modifier.size(if (textAlignment == TabTextAlignment.NO_TEXT) 28.dp else 24.dp)
+            modifier = Modifier
+                        .semantics {
+                            invisibleToUser()
```
joyeeta26, line 126: why are we making this invisibleToUser?
And also this API is deprecated, refer this https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/package-summary#(androidx.compose.ui.semantics.SemanticsPropertyReceiver).invisibleToUser()

fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
```diff
@@ -19,6 +22,7 @@ import com.microsoft.fluentui.tokenized.tabItem.TabItem
 
 data class TabData(
     var title: String,
+    var talkbackDescription: String? = null,  //Custom announcement for Talkback
```
joyeeta26, line 25: rename this to something generic as this will be used by all the screen readers and not specifically talkback

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
Dhruv-Mishra: Thanks for the suggestion, will update the API. 
This does not hide it from the user view. Everything will be visible as is, the UI will be same. It instead ignores it in the accessibility talkback announcement which allows us to override with our own custom talkback announcements only.

fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
Dhruv-Mishra: Updating to 'accessibilityDescription'

fluentui_listitem/src/main/java/com/microsoft/fluentui/tokenized/tabItem/TabItem.kt
Dhruv-Mishra: This is deprecated very recently as part of the latest compose.ui 1.8.0-alpha06 build.

https://developer.android.com/jetpack/androidx/releases/compose-ui#1.8.0-alpha06

The latest stable release is 1.7.5 which does not support this. So we cannot implement the newer method at the moment. Will change the implementation when we upgrade to stable 1.8.0 or higher.

fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
```diff
@@ -19,6 +19,7 @@ import com.microsoft.fluentui.tokenized.tabItem.TabItem
 
 data class TabData(
     var title: String,
+    var accessibilityDescription: String? = null,  //Custom announcement for Talkback
```
joyeeta26, line 22: Please move this optional parameter to the end

fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
```diff
@@ -65,9 +66,16 @@ fun TabBar(
         ) {
             tabDataList.forEachIndexed { index, tabData ->
                 tabData.selected = index == selectedIndex
+                var accessibilityDescriptionValue = if(tabData.accessibilityDescription != null) { tabData.accessibilityDescription }
+                                               else{ tabData.title + if(tabData.selected) ": Active" else ": Inactive" }
```
joyeeta26, line 70: can we add this strings active and inactive to the strings file, so that they are localized correctly

fluentui_tablayout/src/main/java/com/microsoft/fluentui/tokenized/navigation/TabBar.kt
Dhruv-Mishra: Moved

joyeeta26 (approved):

Dhoot (approved):

PR #776, opened 2025-04-29
[V2 Snackbar] swipe to dismiss implementation

joyeeta26:
### Problem 
Implement swipe to dismiss for snackbar

### Screenshots
swipe to dismiss


Dhruv-Mishra:
How does behave with other UI elements like BottomSheets and Drawers within the UI? Would it still work fine?

Dhruv-Mishra (commented):
How does behave with other UI elements like BottomSheets and Drawers within the UI? Would it still work fine?

Dhruv-Mishra (approved):

